PR #6, opened 2021-06-14
Fix typos

dthaler:
Signed-off-by: Dave Thaler <dthaler@microsoft.com>

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -846,7 +846,7 @@ The language datatype represents a human language or family of related languages
 It consists of a sequence of language subtags separated by hyphens,
 where language subtags are [registered by the IANA](https://www.iana.org/assignments/language-subtag-registry).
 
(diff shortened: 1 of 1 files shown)


tychonievich:
- replaces Standard with Specification in `specification/gedcom.md` and `NOTICE`
- replaces old ietf.org links with new rfc-editor links
- removes references to IETF, which were inconsistent in usage, referred to the IETF as an author instead of an organization, conflated RFCs and STDs, and were not mirrored in references to IANA and W3C.

tychonievich (approved):

dthaler:
> * replaces Standard with Specification in `specification/gedcom.md` and `NOTICE`
> * replaces old ietf.org links with new rfc-editor links
> * removes references to IETF, which were inconsistent in usage, referred to the IETF as an author instead of an organization, conflated RFCs and STDs, and were not mirrored in references to IANA and W3C.

@tychonievich  I believe the above comment should have been added to PR #8 instead of this PR.

Issue #12, opened 2021-06-19
Media files called gedcom.ged incompatible with GEDZIP format


fisharebest:
As previously mentioned in https://groups.google.com/g/gedcomgeneral/c/Hh05Hj8X8Ac

If a GEDCOM file contains a media file named `gedcom.ged` (as per the following example), and the data is then saved in GEDZIP format, then the media file will have the same name as the GEDCOM file (`gedcom.ged`).  Clearly we cannot have two files with the same name.

```
0 @X1@ OBJE
1 FILE gedcom.ged
1 NOTE A family tree supplied by my cousin
```

Originally, I suggested that media is stored in a subfolder in the GEDZIP file.  However, now that 7.0 is issued, this will lead to a significant backwards incompatibility, and I suspect this change is not likely to be approved.

As an alternative, we could state that media files MUST NOT have the name `gedcom.ged`.

tychonievich:
I agree this is a problem with the current spec. As you note, the spec has been released so I thing we ban `gedcom.ged` as a `FILE` payload. I also think it worthwhile to ban `MANIFEST.MF` and all paths beginning `META-INF/`; I see that as the intent of the paragraph mentioning them in the current specification.

Issue #11, opened 2021-06-19
Terminology contradiction in FILE tag use of "URL"


dthaler:
The GEDCOM spec says, regarding the FILE tag:

> Syntactically, the payload is a URL, as defined by RFC 3986 and https://url.spec.whatwg.org/. 

and

> A URL with all of the following:
> * no scheme
> ...

However a string with no scheme is not a legal "URL" in either RFC 3986 or https://url.spec.whatwg.org/.
[RFC 3986 sections 4.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-4.1) and 4.2 refer to it as a "relative reference", and 4.1 defines "URI reference" as the term for the union of URIs (including URLs as it explains earlier in the doc) and relative references, and so a "URI reference" would be the RFC 3986 term for what FILE can contain.

The [WHATWG URL spec section 4](https://url.spec.whatwg.org/#urls) similarly defines a URL as the output of a URL parser, and always has a scheme.  The input is a "[valid URL string](https://url.spec.whatwg.org/#valid-url-string)" (equivalent to what RFC 3986 calls a "URI reference") that can be a relative URL (equivalent to RFC 3986's "relative reference") or an absolute URL.

Hence the sentence
> Syntactically, the payload is a URL, as defined by RFC 3986 and https://url.spec.whatwg.org/. 

implies that a scheme is required, and hence contradicts the other text following it.

Presumably the latter text is intended to be correct and the above sentence is incorrect.  One approach would be to change it to something like:

> Syntactically, the payload is a URI reference as defined by RFC 3986, or a valid URL string as defined by https://url.spec.whatwg.org/.   That is, it can be an absolute or relative URL, optionally with a fragment string.

PR #14, opened 2021-06-22
clarify FILE payloads per issues #11 and #12

tychonievich:
Issue #11 noted an inconsistency in the use of the term "URL"
Issue #12 noted an omission to prohibit FILE payloads incompatible with GEDZIP
This pull request attempts to resolve both issues.


tychonievich:
Discussed 2021-06-22
Saying we *must* not use `gedcom.ged` can be seen as making some file that was valid under 7.0.0 invalid under 7.0.4, which technically is a breaking change. However, it is such a corner case it might be acceptable as an error we are correcting in a patch. We'll try drafting a version that only happens in GEDZIP, not GEDCOM, and discuss again.

dthaler:
@tychonievich proposed change looks fine, thanks for addressing #11

dthaler (approved):

tychonievich:
This pull request violates semantic versioning and has been dropped in favor of #16, which does not.

Issue #13, opened 2021-06-22
Clarify extended-use standard structures


tychonievich:
The current specification defines

> An **extended-use standard structure** is a structure whose type, tag, and meaning are defined in this document and whose superstructure is a tagged extension structure.

and recommends

> It is recommended that extensions use extended-use standard structures instead of tagged extension structures if extended-use standard structures will suffice.

However, it does not clarify how to disambiguate these extended-use substructures. For example, should `_XYZ`.`DATE` be interpreted as a `g7:DATE`, `g7:DATE-exact`, `g7:HEAD-DATE`, `g7:NO-DATE`, or `g7:DATA-EVEN-DATE`?

Comparing the current specification with the last pre-URI draft, I believe the following is the intended meaning:

- The URI of a standard tag is determined by the combination of (a) the specification, (b) the superstructure's structure type, and (c) it's tag.
- If the superstructure's structure type is one defined in the specification then, regardless of its tag, it's mapping between substructure tags and substructure structure types is given in the specification.
- Otherwise, if there is a URI defined in the spec as `g7:`*tag*, that is the URI of the structure type of this extended-use standard structure.

For example, that rule would interpret the following example:

```gedcom
0 @1@ _LOC
1 NAME New Amsterdam
2 DATE FROM 1609 TO 1664
1 TYPE Trading post
```

as the following structure hierarchy:

- a tagged extension structure
    - a `g7:NAME`
        - an error; `g7:NAME` has no `DATE`-tagged substructures
    - a `g7:TYPE`

and the following example

```gedcom
0 HEAD
1 GEDC
2 VERS 7.0
1 SCHMA
2 TAG _CHAN https://gedcom.io/terms/v7/CHAN
0 @I1@ INDI
1 NAME Mary //
2 _CHAN
3 DATE JUN 2021
```

as an error because the `DATE` that is a substructure of `g7:CHAN` is `g7:DATE-exact` and the given date lacks a day of the month.

Is this the behavior we believe the 7.0 specification implies? If not, how should it be adjusted? If it is, we'll need to consider how to make this clear in the specification itself.

tychonievich:
Discussed 2021-06-22
Alternative views:
- When you use a standard tag under an extension tag, it must match the semantic meaning and payload rules of at least one standard structure type with that tag, but may have distinct substructures and other limitations imposed by the extension.
or
- This reveals an inconsistency in the "extended-use standard structure" concept, so it should be removed
or
[...]

PR #15, opened 2021-06-22
preliminary draft to address #13

tychonievich:
In an effort to resolve issue #13, adds a new subsection under 1.5 "Extensions" giving a process for identifying structure types.

The text is sufficiently complicated that I plan to let this sit and ponder it for a week or two before moving forward. I welcome other's feedback.

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -592,7 +592,7 @@ If the structure is a record with no superstructure, or if its superstructure ty
 
 If the structure's superstructure type is not defined by this specification and this specification does define a structure type whose URI is the concatenation of `https://gedcom.io/terms/v7/` and the tag, then the structure is an extended-use standard structure and its structure type is given by that URI.
 
-Otherwise, no structure with that tag is permitted in that context and the dataset is invalid.
+Otherwise, no structure with that tag is permitted in that context and the dataset is invalid (but may be valid under a future minor version; see [A Guide to Version Numbers] for more).
 
 :::example
 Consider the following dataset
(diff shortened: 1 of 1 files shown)


specification/gedcom.md
```diff
@@ -580,6 +580,65 @@ In particular, those supporting extensions should keep in mind the following:
...
+
+If a structure's tag is an undocumented extension tag, its structure type is defined by that tag.
+
+If a structure's tag is a standard tag, its structure type depends on its superstructure type as follows:
+
+If the structure is a record with no superstructure, or if its superstructure type is defined in this specification; and if this specification lists a structure with that superstructure and that tag; then the structure type is given by this specification.
+
+If the structure's superstructure type is not defined by this specification and this specification does define a structure type whose URI is the concatenation of `https://gedcom.io/terms/v7/` and the tag, then the structure is an extended-use standard structure and its structure type is given by that URI.
+
+Otherwise, no structure with that tag is permitted in that context and the dataset is invalid (but may be valid under a future minor version; see [A Guide to Version Numbers] for more).
```
dthaler, line 595:
```suggestion
Otherwise, no structure with that tag is permitted in that context and the dataset is invalid (but may be valid under a future minor version; see [A Guide to Version Numbers](#a-guide-to-version-numbers-unnumbered) for more).
```

tychonievich: Because github and pandoc handle internal hyperlinks differently, I'd rather leave it as just `[A Guide to Version Numbers]`. We could use `a-guide-to-version-numbers` or add a custom section ID and it would work on gedcom.io but be a dead link in github; or add `a-guide-to-version-numbers-unnumbered` and it would be a dead link on gedcom.io but work in github; but I think the simple brackets are about as good as we can get in trying to make both readable.

tychonievich:
This draft has too many omissions and has been rejected in preference of #19

PR #16, opened 2021-06-22
clarify FILE and GEDZIP to resolve #11 and #12

### LINKED ISSUE #11: Terminology contradiction in FILE tag use of "URL"

tychonievich:
Alternative to #14 that puts a recommendation, not requirement in the FILE structure and moves the requirement to GEDZIP


specification/gedcom.md
```diff
@@ -176,8 +176,8 @@ Use of Unicode-aware processing and display libraries is recommended.
 
 Character-level grammars are specified in this document using
 Augmented Bakaus-Naur Form (ABNF)
-as defined in [STD 68](https://www.rfc-editor.org/info/std68)
-and modified in [RFC 7405](https://www.rfc-editor.org/info/rfc7405).
+as defined in STD 68 (<https://www.rfc-editor.org/info/std68>)
+and modified in RFC 7405 (<https://www.rfc-editor.org/info/rfc7405>).
```
dthaler, line 180: This regresses PR #9.  The original is more correct so recommend reverting this change.

specification/gedcom.md
```diff
@@ -898,7 +898,7 @@ Lineage-linked data pertains to individuals linked in family relationships acros
 The genealogical structures defined in this chapter are based on the general framework of the container format and data types defined in Chapters 1 and 2.
 
 Historically, these genealogical structures were used as the only form approved for exchanging data with Ancestral File, TempleReady and other Family History resource files.
-Those systems were all replaced between 1999 and 2019, and [GEDCOM-X](https://gedcomx.org) was introduced as the new syntax for communication with their replacements.
+Those systems were all replaced between 1999 and 2019, and GEDCOM-X (<https://gedcomx.org>) was introduced as the new syntax for communication with their replacements.
```
dthaler, line 901: Regression since this was fixed in PR #9

specification/gedcom.md
```diff
@@ -3737,14 +3741,15 @@ It is often useful to transmit a dataset together with a set of external files.
 The FamilySearch GEDZIP 7.0 file format is provided for this purpose.
 Version 7.0 was the first version of GEDZIP released; the version number of a GEDZIP file is the same as the version number of the dataset it contains.
 
-A GEDZIP file is a zip archive, as defined by the [.ZIP File Format Specification](http://www.pkware.com/appnote)
+A GEDZIP file is a zip archive, as defined by <http://www.pkware.com/appnote>
```
dthaler, line 3744: Regression from fix in PR #9

dthaler (changes requested):

PR #19, opened 2021-07-06
new sections on extensions per #13

tychonievich:
Quite a lot of new text because I wanted to be clear and cover every case. Not all of the cases have been discussed previously, though; a few we should discuss (and possibly revise) before merging include:

- Can an tagged extension structure use standard enumerated values? I thought yes
    - And I picked URI prefixes as the first of the following that works: `g7:enum-`, `g7:`, `g7:cal-`. `g7:month-`; is that right?

- Can an extended-use standard structure use non-standard enumerated values? I thought no

PR #24, opened 2021-07-16
draft that removes under-specified extended-use standard structures

tychonievich:
A solution to #13 using a different tack than #19:

Given we had defined a term "extended-use standard structures" but had not provided semantics for determining their meaning, this PR simply removes them entirely.

This PR also clarifies that
- `_LOC`.`NAME` is OK, and is defined in the specification for `_LOC`
- `_LOC`.`NAME._XYZ` is OK, and is defined in the specification for `_XYZ` not `_LOC`
- If we have `2 TAG _DATE https://gedcom.io/terms/v7/DATE` then we can use `_DATE` iff we can't use `DATE` to mean a `g7:DATE`.

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -402,7 +402,7 @@ as described in [Extensions].
 
 A **standard structure** is a structure whose type, tag, meaning, superstructure, and cardinality within the superstructure are described in this document. This includes records such as `INDI` and substructures such as `INDI`.`NAME`.
 
-The *recommended* way to go beyond the set of standard structure types in this specification or to expand their usage is to submit a feature request on the [FamilySearch GEDCOM development page](https://github.com/FamilySearch/GEDCOM/issues) so that the ramifications of the proposed addition and its interplay with other proposals may be discussed and the addition may be included in a subsequent version of this specification.
+The recommended way to go beyond the set of standard structure types in this specification or to expand their usage is to submit a feature request on the [FamilySearch GEDCOM development page](https://github.com/FamilySearch/GEDCOM/issues) so that the ramifications of the proposed addition and its interplay with other proposals may be discussed and the addition may be included in a subsequent version of this specification.
 
 This specification also provides multiple ways to for extension authors to go beyond the specification without submitting a feature request, which are described in the remainder of this section.
 
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -1,6 +1,6 @@
 ---
 title: The FamilySearch GEDCOM Specification
-subtitle: 7.0.3
+subtitle: 7.0.4
 email: GEDCOM@ChurchOfJesusChrist.org
 copyright: |
     :::{style="page-break-after: always;page-break-before: always;"}
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -402,6 +402,9 @@ as described in [Extensions].
 
 A **standard structure** is a structure whose type, tag, meaning, superstructure, and cardinality within the superstructure are described in this document. This includes records such as `INDI` and substructures such as `INDI`.`NAME`.
 
+The recommended way to go beyond the set of standard structure types in this specification or to expand their usage is to submit a feature request on the [FamilySearch GEDCOM development page](https://github.com/FamilySearch/GEDCOM/issues) so that the ramifications of the proposed addition and its interplay with other proposals may be discussed and the addition may be included in a subsequent version of this specification.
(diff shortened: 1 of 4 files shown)


specification/gedcom.md
```diff
@@ -402,16 +402,63 @@ as described in [Extensions].
...
+1 NAME Βυζάντιον
+2 DATE FROM 667 BCE TO 324
+1 _POP 15149358
+2 DATE 31 DEC 2020
+0 @I1@ INDI
+1 BIRT
+2 _LOC @P1@
+```
+
+- Both `_LOC` are tagged extension structures, as is `_POP`.
```
dthaler, line 425:
```suggestion
- Both uses of `_LOC` are tagged extension structures, as is `_POP`.
```

specification/gedcom.md
```diff
@@ -402,16 +402,63 @@ as described in [Extensions].
...
+- `_LOC`.`NAME` and `_LOC`.`NAME`.`DATE` are both extension-defined substructures. Their meaning is defined by the specification defining `_LOC`.
+- `_POP`.`DATE` is an extension-defined substructure. Its meaning is defined by the specification defining `_POP`.
+- Even though both `DATE`s appear to have `g7:type-DATE` payloads, we can't know that is the intended datatype without consulting the defining specifications of `_LOC` and `_POP`, respectively.
```
dthaler, line 428: My view as an implementer is this is not good for implementation.   I'd strongly prefer to keep standard names as having the datatypes in the spec.

tychonievich:
I've love feedback on the new "requirements and recommendations" section, particularly the new items:

>- It is required that each extension structure type use the same payload type in each place it appears.

Should this be a recommendation instead?

>- In order from most-preferred to least-preferred, the options for extensions are:
>    1. A feature request for inclusion in a future version of this specification (and then waiting for that version's release).
>    2. A relocated standard structure with a documented extension tag.
>    3. A tagged extension structure, calendar, or enumerated value with a documented extension tag.
>    4. An extension-defined substructure beneath a documented extension tag.
>    5. A tagged extension structure, calendar, or enumerated value with a undocumented extension tag.
>    6. An extension-defined substructure beneath an undocumented extension tag.

Is this the right order?
Do we want to express it like this? We could also do it as a set of "you shouldn't use X if Y will suffice" recommendations instead.

specification/gedcom.md
tychonievich: Good point. Example text updated.

specification/gedcom.md
```diff
@@ -402,16 +402,67 @@ as described in [Extensions].
...
+:::
+
+If an extension-defined substructure has a tag that is also used by one or more standard structures, its meaning and payload type should match at least one of those standard structure types.
+
+:::example
+An extension-defined substructure with tag "`DATE`" should provide a date or date period relevant to its superstructure, as do all `DATE`-tagged structures in this specification. Extensions should not use "`DATE`" to tag a structure describing anything else (even something that might reasonably be abbreviated "date", such as someone an individual dated).
+:::
+
+As a special case, a tagged extension structure can be defined to have a standard structure type.
+These are called **relocated standard structures** and can only appear with superstructures that are not documented as a superstructure of that structure type in this specification.
```
dthaler, line 442: Is there a technical reason for this limitation?
This might prevent arguably legitimate cases, such as DATE.TIME with local time, and DATE._TIME with UTC (say because an obit stated time of death, but the time zone boundary or DST rule was updated in a later year so listing the UTC disambiguates which time zone it was in at the time).   With this one would have to define _TIME to use something other than g7:type-Time even if the syntax is the same as g7:type-Time.   I just made up this use case right now and I don't need it, there may be other legit ones that people do need.

tychonievich: `g7:type-Time` is a payload type, not a structure type, and its not limited by this clause. It's perfectly OK to define a structure type _UTCTIME with payload type g7:type-Time and use it next to a g7:TIME structure type. But `g7:TIME` is a structure type and is limited by this clause.

Why the limit?
- `g7:TIME` should be a full description of the structure type. It shouldn't have variants for `TIME` vs `_TIME` as the tag.
- Structure types define semantics, and singular structures define semantics that don't generalize to plurality. Is a DATE with two TIMEs suppose to be an event that happened twice in a day, or under uncertainty about the time, or the same time in two time zones, or ...?

Result: if there's already a g7:TIME substructure, you can't define a new tag for it. Either it's already plural, in which case use several with tag TIME, or you have some new plural semantics in mind and should define a structure type to explain that.

PR #51, opened 2021-09-10
Line-based (not byte-based) alternative

tychonievich:
I drafted this to see what it would look like. If we like it, we probably need to say something about character encoding too.
I also added all the 3rd-party specs I know of.

### DIFF
--- a/version-detection/version-detection.md
+++ b/version-detection/version-detection.md
@@ -21,7 +21,7 @@ All version of GEDCOM files are text files that conform to the following general
     - The first line's level is 0.
     - Each other line's level is at most on greater than the the level of the line before it.
   2. An optional **identifier**, which begins and ends with a COMMERCIAL AT (U+0040 `@`) and contains no internal line terminators or COMMERCIAL ATs. No two lines share the same identifier.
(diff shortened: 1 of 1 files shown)


version-detection/version-detection.md
```diff
@@ -11,87 +11,61 @@ to parse the data.  The GEDCOM format itself does include a version identifier i
...
-Initial bytes (hex) | Width | Order
-------------------- | ----- | -------------
-FF FE               | 2     | LE (Little-endian)
-30 00               | 2     | LE (Little-endian)
-FE FF               | 2     | BE (Big-endian)
-00 30               | 2     | BE (Big-endian)
-Otherwise           | 1     | (N/A)
 
-## Version Detection
+## Version detection
```
dthaler, line 33: I have no problem with the _addition_ of the new text above, as long as it is informative not normative.
However, the text removal above and changes below are, in my view, problematic as it lost all discussion of character set.  One cannot do string matching without a discussion of character set.  As such I'd prefer to keep byte matching, even if informative text such as above is added as an introduction.

tychonievich:
Discussed 2021-09-14: decided not to do this as it lacks character points

PR #53, opened 2021-09-12
Use consistent shared note terminology

dthaler:
use "shared note" (like everywhere else in the doc), not "sharable note"

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -1329,7 +1329,7 @@ The origin of a name might be a reasonable shared note, while the reason a parti
 :::
 
 :::note
-Although sharable notes have been present since version 5.0 was released in 1991, as of 2021 relatively few applications have a user interface that presents shared notes as such to users. It is recommended that `SNOTE` be avoided when `NOTE` will suffice.
+Although shared notes have been present since version 5.0 was released in 1991, as of 2021 relatively few applications have a user interface that presents shared notes as such to users. It is recommended that `SNOTE` be avoided when `NOTE` will suffice.
 :::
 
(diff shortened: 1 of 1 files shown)


tychonievich:
I used a different term here intentionally, as I didn't want to be misinterpreted to mean that `SNOTE` has been in GEDCOM; rather, `NOTE` could be used like both 7.0's `NOTE` and like 7.0's `SNOTE`. But I see how the current wording is confusing.

I'm OK with this proposed change, but wonder if the following would be better?

> The ability to have multiple structures share a single note using pointers was introduced in version 5.0 in 1991. However, as of 2021 relatively few applications have a user interface that presents shared notes as such to users. It is recommended that `SNOTE` be avoided when `NOTE` will suffice.

fisharebest:
> It is recommended that SNOTE be avoided when NOTE will suffice.

Why recommend this?

Shared notes are part of the standard, and any application that reads GEDCOM 7 files should read them.

Now, if the GEDCOM-reader doesn't expose them in its user-interface, then it may well do something such as:
* convert each `SNOTE` into multple copies of the same `NOTE`
* display them, but keep them read-only
* ignore them

But that shouldn't be the concern of the GEDCOM-writer.

(Disclaimer - my own application makes extensive use of shared-notes and exposes them via the GUI.)

tychonievich:
> Why recommend this?

When we have two ways to do one thing (such as a single-use `SNOTE` vs a `NOTE`), we tried to provide a recommendation on how to chose between them. When the two are equally expressive, we generally recommended the more widely-supported option.

In this particular case, we had many stakeholders asking us to remove pointers to notes entirely. We rejected that request because we knew of several applications (including yours) that use them to good effect.

dthaler:
Updated PR per @tychonievich suggestion

PR #56, opened 2021-09-16
Fix head lang

tychonievich:
Clarifies that `g7:HEAD-LANG` should only be used if *all* Text-type payloads with no `g7:LANG` substructure are in the same language.


tychonievich:
mistakenly included too many files

Issue #55, opened 2021-09-16
HEAD.LANG and untagged payloads


tychonievich:
5.5.1 defined HEAD.LANG as "The human language in which the data in the transmission is normally read or written.  It is used primarily by programs to select language-specific sorting sequences and phonetic name matching algorithms."

7.0 clarified that as "a default value for the rest of the document" and "The language in which the `Text`-typed payloads of all structures in the document appear, except as superseded by a `g7:LANG`."

However, neither is accurate in practice. There are many `<Text>`-payload structures that do not admit a `g7:LANG` substructure, and there is no reason to believe they all have the same human language. As written, merging files with different HEAD.LANG is impossible.

In the 2021-09-14 steering committee meeting, this was identified as an error in the spec, and one to be fixed by a patch as soon as possible. This is a patch under the provision "If there is an ambiguity or contradiction in the specification, it will be resolved in a patch version unless it is known that implementations interpreted the spec differently and that clarifying the intended meaning would cause incompatibilities between those implementations."

tychonievich:
Options:
1. We could keep the current text but say the language tag `und` should be used unless all `<Text>` payloads without a `g7:LANG` substructure are in the same language
2. Redefine `g7:HEAD-LANG` to be something vague like a "language hint" instead of the more formal language of untagged `<Text>` payloads
3. Add some kind of placeholder text like "This structure indicates a language used in some subset of the rest of the dataset. However, the current specification does not have enough data to fully identify the language of all Text payloads. Once a future version defines how to specify the language of every Text payload, `g7:HEAD-LANG` will be a default language tag.

Preferences? Other options?

fisharebest:
I prefer 1.

Since `HEAD.LANG` is optional, perhaps state that if it missing, then a GEDCOM-reader should assume `und`.

dthaler:
Albert's technical points convinced me that option 2 is the best path, since "unless all <Text> payloads without a g7:LANG substructure are in the same language" is not implementable.

[RFC 6365](https://datatracker.ietf.org/doc/html/rfc6365#page-29) says:

>    language identification
> 
>       Specification of the human language used for a string of text.
>       <RFC6365>
> 
>       Some protocols (such as MIME and HTTP) allow text that is meant
>       for machine processing to be identified with the language used in
>       the text.  Such identification is important for machine processing
>       of the text, such as by systems that render the text by speaking
>       it.  Language identification is also called "language tagging".
[...]

PR #57, opened 2021-09-16
Clarify g7:HEAD-LANG per #55 option 1

tychonievich:
No change in spec, but significant new clarification of g7:HEAD-LANG per issue #55 option 1


specification/gedcom.md
```diff
@@ -2831,10 +2831,21 @@ A [registry of component subtags](https://www.iana.org/assignments/language-subt
 
 #### `LANG` (Language) `g7:HEAD-LANG`
 
-The language in which the `Text`-typed payloads of all structures in the document appear, except as superseded by a `g7:LANG`.
+The language in which the `Text`-typed payloads of all structures in the dataset appear, except as superseded by a `g7:LANG`.
```
dthaler, line 2834: I agree with Albert's technical points raised during the last meeting, in that I don't believe this sentence is implementable in general.   A user can type in text from some language and the program cannot guarantee it's in the language the program thinks it is, without actually asking the user (who may then give the wrong answer).

tychonievich:
Discussed 2021-09-21; decided this was the wrong approach

PR #59, opened 2021-09-21
Default language meaning (Option 2)

dthaler:
Addresses issue #55 

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -2831,7 +2831,14 @@ A [registry of component subtags](https://www.iana.org/assignments/language-subt
 
 #### `LANG` (Language) `g7:HEAD-LANG`
 
-The language in which the `Text`-typed payloads of all structures in the document appear, except as superseded by a `g7:LANG`.
+The default language using which to interpert any `Text`-typed payloads of all structures without their own
+`g7:LANG` substructure.  Language information is often used for purposes such as language-specific sorting
+sequences and phonetic, name matching algorithms, spell-checking,
+computer-synthesized speech, Braille transcription, or language translation, and most operating
+systems allow configuring a language via application language settings or "user locales".  Languages
+settings can also be set by Input Method Editors (IMEs), and it is expected that applications will
+set the default language based on the current language in use by the IME, user locale, or application
+language setting, as appropriate, rather than being any guarantee based on the content of actual strings.
 
 The payload of the `LANG` structure is a language tag, as defined by [BCP 47](https://www.rfc-editor.org/info/bcp47).
 


specification/gedcom.md
```diff
@@ -2831,7 +2831,14 @@ A [registry of component subtags](https://www.iana.org/assignments/language-subt
 
 #### `LANG` (Language) `g7:HEAD-LANG`
 
-The language in which the `Text`-typed payloads of all structures in the document appear, except as superseded by a `g7:LANG`.
+The default language using which to interpert any `Text`-typed payloads of all structures without their own
+`g7:LANG` substructure.  Language information is often used for purposes such as language-specific sorting
```
tychonievich, line 2835: "`g7:LANG` substructure" is too specific; `g7:LANG` can apply to peer structures and their substructures, as e.g. `NAME`.`LANG` applying to `NAME`.`GIVN`. This is why I used "superseded by a `g7:LANG` in the previous text.

specification/gedcom.md
```diff
@@ -2831,7 +2831,14 @@ A [registry of component subtags](https://www.iana.org/assignments/language-subt
...
+sequences and phonetic, name matching algorithms, spell-checking,
+computer-synthesized speech, Braille transcription, or language translation, and most operating
+systems allow configuring a language via application language settings or "user locales".  Languages
```
tychonievich, line 2838: Because these are **user** locales, shouldn't the first `SUBM`.`LANG` of the `SUBM` record describing the current user be used in lieu of `HEAD`.`LANG` if available? Or am I misunderstanding the meaning of "user" here?

tychonievich:
Mulling over how to fix the substructure point and how to separate the normative definition of `g7:HEAD-LANG` from the informative contextual information, how about the following?

> A default language which may be used to interpret any `Text`-typed payloads that lack a specific language tag, as given by `g7:LANG`. Applications may also select a different default based on its knowledge of the application user.
>
> :::note
> Not all `Text`-type payloads have a language tag, but language information may be needed for purposes such as language-specific sorting sequences, phonetic name matching algorithms, spell-checking, computer-synthesized speech, Braille transcription, and language translation. `g7:HEAD-LANG` is a default language for these purposes supplied by the system creating the dataset, and may be superseded by operating system settings, user locales, Input Method Editors (IMEs), and other means of language specification.
>
> When an explicit language tag for a given `Text`-typed payload is available, it should supercede any default language.
> :::

In writing the above I realized I don't know what priority to suggest between `g7:HEAD-LANG`, IME settings, and user locales.

dthaler:
> In writing the above I realized I don't know what priority to suggest between `g7:HEAD-LANG`, IME settings, and user locales.

Typically the priority order is, I _think_: app internal override, IME, app-specific setting in OS, user locale, system locale.

dthaler:
And I imagine often the desired behavior between what-the-GEDCOM-says vs what the local setting is, might vary by algorithm / use case.  For example, an app might want name matching to be based on the language of the name, whereas the sorting algorithm (for all names regardless of language, like in a selection box or report) to be what the viewing user is used to.

tychonievich:
Pulled #62 instead of #59

PR #61, opened 2021-09-21
Add links to gedcom.io supporting material

tychonievich:


tychonievich:
I'm not sure if this is the right level of detail or not. Is the bulleted list of example resources appropriate here?

dthaler (approved):

PR #62, opened 2021-09-23
HEAD-LANG option 2 (hint) rewording

tychonievich:
Possible solution of #55
Same intent as #57 but differently expressed


tychonievich:
Discussed and approved in steering committee meeting 2021-09-28

PR #79, opened 2021-11-02
Add proposed definition of the GEDZIP media type

### LINKED ISSUE #5: Media types for GEDCOM and GEDZIP
["application/zip"](https://www.iana.org/assignments/media-types/application/zip) is a media type for ZIP files in general.

However, it is common practice when there are additional constraints specified, such as GEDZIP does, to define a new media type.  For one of many examples see [application/epub+zip](https://www.iana.org/assignments/media-types/application/epub+zip) and search for zip in the [media types registry](https://www.iana.org/assignments/media-types/media-types.xhtml) for many other examples).

Especially since GEDZIP defines a new file extension `.gdz` instead of simply reusing `.zip` then for the same reason it's just as important to register a more specific media type for GEDZIP and have it point to the Gedcom7 spec as the specification for it.

dthaler:
Since it was decided that the GEDZIP version and gedcom.ged file version
must match, then the version parameter in the media type can be used to
specify how to interpret the GEDZIP format.

But since .gdz files come without a media type, we still need to specify
how to interpret a .gdz file when you don't know if it's 7.0 or some
future version.

a) We require gedcom.ged to always be in the same place with the same name in all future GEDZIP versions, or
b) We would need to specify a version detection algorithm for GEDZIP if that ever changed.

The choice of a vs b does affect the wording in the existing media type request's version parameter, so we should probably choose now.

Fixes #5

### DIFF
--- a/version-detection/version-detection.md
+++ b/version-detection/version-detection.md
@@ -114,6 +114,8 @@ Byte sequence  | Explanation | Reference
 
 ## IANA Media Type Registration
 
+### text/vnd.familysearch.gedcom
+
 This section defines the "text/vnd.familysearch.gedcom" media type for use with protocols such as HTTP or email, to identify content
 as being a GEDCOM file.  The Internet Assigned Numbers Authority (IANA) maintains the
 [registry of assigned media types](https://www.iana.org/assignments/media-types/media-types.xhtml).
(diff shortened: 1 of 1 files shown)


tychonievich:
We can't provide a version detection algorithm yet, as we don't know how a possible future version would be detected. Thus, we'll need to redo the registration if/when a version that doesn't use `gedcom.ged` comes along, won't we? If so, I think we should use the simpler wording that assumes it's always `gedcom.ged` and that the version parameter is optional for now.

tychonievich (approved):

dthaler:
Merging per meeting discussion

Issue #91, opened 2021-11-23
Trailing spaces and empty payloads


tychonievich:
This issue originally came up in #81

We have

> ```abnf
> Line = Level D [Xref D] Tag [D LineVal] EOL
> ```

which would mean that 

- an empty-string `NOTE` would be encoded "`1 NOTE `" with a trailing space
- a no-payload `CHAN` would be "`1 CHAN`" without a trailing space
- a no-payload `MARR` would be "`1 MARR`" without a trailing space, because of the metasyntax definition 
  > `[`text`|<NULL>]` means the payload is optional but if present must be the given text.
  
  i.e., the empty string is not a valid payload for `[Y|<NULL>]` but a missing payload is

We also have 

> Empty payloads and missing payloads are considered equivalent. The remainder of this document uses “payload” as shorthand for “non-empty payload”.

which suggests that this distinction between no-payload and empty-payload is not meaningful, despite being present in the spec.

The question raised, but not answered, in #81 is (a) should the spec provide a recommendation on the use of trailing spaces with empty payloads and (b) if so, what should it be?

dthaler:
Hmm.  The spec says:

```
Line = Level D [Xref D] Tag [D LineVal] EOL
LineVal = pointer / lineStr
nonAt = %x09 / %x20-3F / %x41-10FFFF ; non-EOL, non-@
lineStr = (nonAt / atsign atsign) *nonEOL ; leading @ doubled
```

So the ABNF does not permit LineVal to be empty, the space must be followed by an atsign or nonAt character.

tychonievich:
Need to clarify that the way to encode the empty string payload is to not have a `LineVal`

PR #93, opened 2021-11-23
Trailing whitespace discouraged

tychonievich:
One possible solution to #91

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -325,7 +325,9 @@ are encoded with no `LineVal` and no space after the `Tag`.
 :::example
 The payload of a `MARR` structure has type `[Y|<NULL>]`, which is optional but cannot be the empty string.
 The payload of a `EVEN` structure has type `Text`, and `Text` can be the empty string.
-Both no-payload "`1 MARR`" and empty-payload "`1 EVEN`" are represented with no `LineVal` and no trailing space.
+The `Line` encoding a no-payload `MARR` is "`1 MARR`"
+and the `Line` encoding an empty-payload `EVEN` is "`1 EVEN`";
+both `Line`s have no `LineVal` and no trailing space.
 :::
 
 If a line value matches production `Xref`, the same value must occur as the cross-reference identifier of a structure within the document.
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -330,9 +330,6 @@ and the `Line` encoding an empty-payload `EVEN` is "`1 EVEN`";
 both `Line`s have no `LineVal` and no trailing space.
 :::
 
-:::note
-Implementations may choose to be more permissive and parse lines ending with a `Tag` and a single space as if they had no `LineVal`, but are not required to do so. They must not create lines of this form.
-:::
 
 If a line value matches production `Xref`, the same value must occur as the cross-reference identifier of a structure within the document.
 The special `voidPtr` production is provided to encode null pointers.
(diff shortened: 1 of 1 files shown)


specification/gedcom.md
```diff
@@ -316,6 +316,20 @@ Each non-pointer payload may be encoded in 1 or more line strings (line continua
 The exact encoding of non-pointer payloads is dependent on the datatype of the payload, as determined by the structure type.
 The datatype of non-pointer payloads cannot be fully determined by line value content alone.
 
+Because empty payloads and missing payloads are considered equivalent,
+both a structure with no payload
+and a structure with the empty string as its payload
+may be encoded either
+with an empty `LineVal` (i.e. with a space after the `Tag`)
+or with no `LineVal` (i.e. without a space after the `Tag`).
```
dthaler, line 324:
```suggestion
with an empty `LineVal` (i.e., with a space after the `Tag`)
or with no `LineVal` (i.e., without a space after the `Tag`).
```

specification/gedcom.md
```diff
@@ -316,6 +316,20 @@ Each non-pointer payload may be encoded in 1 or more line strings (line continua
 The exact encoding of non-pointer payloads is dependent on the datatype of the payload, as determined by the structure type.
 The datatype of non-pointer payloads cannot be fully determined by line value content alone.
 
+Because empty payloads and missing payloads are considered equivalent,
+both a structure with no payload
+and a structure with the empty string as its payload
+may be encoded either
+with an empty `LineVal` (i.e., with a space after the `Tag`)
+or with no `LineVal` (i.e., without a space after the `Tag`).
+In both cases, it is recommended that the space be omitted.
```
tychonievich, line 325:
```suggestion
In both cases, it is recommended that the `LineVal` be omitted.
```

tychonievich: Inconsistent with spec; see #91 which says it's not discouraged, it's prohibited

specification/gedcom.md
```diff
@@ -316,6 +316,24 @@ Each non-pointer payload may be encoded in 1 or more line strings (line continua
...
 The datatype of non-pointer payloads cannot be fully determined by line value content alone.
 
+Note that production `LineVal` does not match the empty string.
+Because empty payloads and missing payloads are considered equivalent,
+both a structure with no payload
+and a structure with the empty string as its payload
+are encoded with no `LineVal` and no space after the `Tag`.
+
+:::example
+The payload of a `MARR` structure has type `[Y|<NULL>]`, which is optional but cannot be the empty string.
```
dthaler, line 326:
```suggestion
The payload of a `MARR` structure has type `[Y|<NULL>]`, which is optional but if present cannot be the empty string.
```

specification/gedcom.md
```diff
@@ -316,6 +316,24 @@ Each non-pointer payload may be encoded in 1 or more line strings (line continua
...
+The payload of a `EVEN` structure has type `Text`, which is not optional but can be the empty string.
+The `Line` encoding a no-payload `MARR` is "`1 MARR`"
+and the `Line` encoding an empty-payload `EVEN` is "`1 EVEN`";
+both `Line`s have no `LineVal` and no trailing space.
+:::
+
+:::note
+Implementations may choose to be more permissive and parse lines ending with a `Tag` and a single space as if they had no `LineVal`, but are not required to do so. They must not create lines of this form.
+:::
```
dthaler, line 335: I'm ok with this note, but I don't think it's necessary and would prefer removing it.

dthaler: Rationale for removal is that there's lots of other ways throughout the spec that implementations may choose to be more permissive and parse lines that should never occur, but we don't have notes for them.

tychonievich:
```suggestion
```

Issue #80, opened 2021-11-07
Modify cardinality of EXID.TYPE


albertemmerich:
EXID.TYPE is {0:1}.
But
1 EXID 123456
without any TYPE does not make sense.
EXID.TYPE should be {1:1}
As this affects the validity of existing 7.0 files, it is a point for GEDCOM 8.

tychonievich:
Good catch! I don't know how we missed that before.
The definition of EXID depends heavily on EXID.TYPE: it asserts every identifier is maintained by an authority given in the TYPE and defers to the TYPE to define uniqueness and other semantic properties of the EXID. That suggest that EXID without TYPE is ambiguous, and thus possibly patchable? I feel like we're in a gray area of semantic versioning there.

tychonievich:
Proposed:
Add to spec a notice that EXID.TYPE is strongly recommended and EXID without TYPE will be removed in the next version of the specification.

PR #95, opened 2021-11-23
Deprecation EXID without TYPE

tychonievich:
As per #80, EXID without TYPE is meaningless and should not be used.

I put it three places: in both `IDENTIFIER_STRUCTURE` and `PLACE_STRUCTURE` (which list the EXID/TYPE relationship) and the description of EXID (so that it will show up in the tag definition served at <https://gedcom.io/terms/v7/EXID>)

I've also added a new "deprecation note" environment to the CSS to make this stand out more in the rendered HTML and PDF.

### DIFF
--- a/specification/gedcom.md
+++ b/specification/gedcom.md
@@ -2675,12 +2675,6 @@ the authority owning the identifier is provided in the TYPE substructure; see `E
 
 Depending on the maintaining authority, an `EXID` may be a unique identifier for the subject, an identifier for 1 of several views of the subject, or an identifier for the externally-maintained copy of the same information as is contained in this structure. However, unlike `UID` and `REFN`, `EXID` does not identify a structure; structures with the same `EXID` may have originated independently rather than by edits from the same starting point.
 
-:::deprecation
-Having an `EXID` without an `EXID`.`TYPE` substructure is deprecated.
-The meaning of an `EXID` depends on its `EXID`.`TYPE`.
-The cardinality of `EXID`.`TYPE` will be changed to `{1:1}` in version 8.0.
-:::
-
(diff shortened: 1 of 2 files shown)


specification/gedcom.md
```diff
@@ -2659,6 +2675,12 @@ the authority owning the identifier is provided in the TYPE substructure; see `E
 
 Depending on the maintaining authority, an `EXID` may be a unique identifier for the subject, an identifier for 1 of several views of the subject, or an identifier for the externally-maintained copy of the same information as is contained in this structure. However, unlike `UID` and `REFN`, `EXID` does not identify a structure; structures with the same `EXID` may have originated independently rather than by edits from the same starting point.
 
+:::deprecation
+Having an `EXID` without an `EXID`.`TYPE` substructure is deprecated.
+The meaning of an `EXID` depends on its `EXID`.`TYPE`.
+The cardinality of `EXID`.`TYPE` will be changed to `{1:1}` in version 8.0.
+:::
```
dthaler, line 2682: I don't think the note is needed here since there's nothing in this section that is otherwise wrong, and line 2673-2674 already implies the TYPE has to exist.  The other two locations look good.

tychonievich: I added it here because this is the section that gets auto-extracted as https://gedcom.io/terms/v7/EXID and I wanted it to show up there. But you're right, it looks odd to have a deprecation in the spec that refers to something not local to it.

dthaler: I would maintain that it's not needed there either because it already says "the authority owning the identifier
**is** provided in the TYPE substructure" (i.e., "is provided" not "may be provided")

tychonievich: Good point. Fixed

Issue #113, opened 2022-02-19
Nonsensical days in DateExact


dthaler:
The spec currently has:
```
date        = [calendar D] [[day D] month D] year [D epoch]
day     = Integer
year    = Integer
month   = stdTag / extTag  ; constrained by calendar
```

Thus "2022 FEB 18" would be considered a legal date, being the 2022'nd day of February in the year 18 AD.   Is that intentionally legal?   To me this would seem better to not treat as legal or it will just encourage dates being entered in the wrong format (2022 FEB 18 instead of 18 FEB 2022).

Similarly "0 FEB 2022" is also currently a legal date it seems, which is also non-sensical.

Options:
1. No change.   "2022 FEB 18" and "0 FEB 2022" are both legal dates.
2. Constrain the day range independent of the calendar, such as to a value in the range [1,36] or even [1,366].
3. Constrain the day range based on a calendar-specific rule.

My preference would be to do 3 if the calendar is recognized, and 2 for unrecognized calendars.

dthaler:
GEDCOM Steering Committee: general agreement to make some change for the DateExact.   Option 2 would be appropriate in all cases (James to find reasonable bounds).  Option 3 is ok to allow implementations to do if they recognize the calendar.

tychonievich:
Discussed in re patch/minor/major. Decided that the admission of invalid dates was an oversight to be patched.

We plan to write up a patch that does something like

1. Ban invalid dates in any given calendar (without defining what that means)
2. State that "across known calendars, none of them use day 0 or have day greater than 36, and we expect others won't either" or the like

elyoh:
> State that "across known calendars, none of them use day 0 or have day greater than 36, and we expect others won't either" or the like

Please can you outline the basis for the upper bound of 36.

dthaler:
> Please can you outline the basis for the upper bound of 36.

There are no calendars we know of that have months of more than 36 days, especially since the lunar cycle is less than that.  https://en.wikipedia.org/wiki/History_of_calendars discusses many historical calendars, of which under China, it says "In the so-called [five-phase](https://en.wikipedia.org/wiki/Wuxing_(Chinese_philosophy)) calendar, the year consists of 10 months and a transition, each month being 36 days long, and the transitions 5 or 6 days." so there is evidence of a calendar with 36 days per month.

PR #125, opened 2022-03-17
Limit days and years by calendar

tychonievich:
Previously we said "the allowable `month`s and `epoch`s are determined by the `calendar`."
Now we say "the allowable `day`s, `month`s, `year`s, and `epoch`s are determined by the `calendar`."
Also a note about the range of days in known calendars.

See #113 for more.

### DIFF
--- a/specification/gedcom-02-datatypes.md
+++ b/specification/gedcom-02-datatypes.md
@@ -88,7 +88,9 @@ epoch   = %s"BCE" / extTag ; constrained by calendar
 
 In addition to the constraints above:
 
-- The allowable `month`s and `epoch`s are determined by the `calendar`.
+- The allowable `day`s, `month`s, `year`s, and `epoch`s are determined by the `calendar`.
+    All known calendars restrict `day` to be between 1 and a month-specific maximum.
+    The largest known maximum is 36, and most months in most calendars have a lower maximum.
 - No calendar names, months, or epochs match `dateRestrict`.
 - Extension calendars (those with `extTag` for their `calendar`) must use `extTag`, not `stdTag`, for months.
 


specification/gedcom-02-datatypes.md
```diff
@@ -88,7 +88,9 @@ epoch   = %s"BCE" / extTag ; constrained by calendar
 
 In addition to the constraints above:
 
-- The allowable `month`s and `epoch`s are determined by the `calendar`.
+- The allowable `day`s, `month`s, `year`s, and `epoch`s are determined by the `calendar`.
+    All known calendars restrict `day` to be between 1 and a month-specific maximum.
+    The largest known maximum is 36, and most months in most calendars have a lower maximum.
```
dthaler, line 93: This text allows for a calendar specific rule, although it would be nice to be more explicit about how to handle an unrecognized calendar.  E.g., "If the calendar is unrecognized, implementations can assume that the day is restricted to a value between 1 and 36."

fisharebest:
> "If the calendar is unrecognized, implementations can assume that the day is restricted to a value between 1 and 36."

Why is this statement necessary?

I don't see the benefit of either having it or trying to enforce it.  What should software do when it encounters such a date?  We already know it it invalid because we don't recognise the calendar.

I guess most systems will just display the raw data.  If the data says `2 DATE MYCALENDAR 37 FOO 1234`, should we display `MYCALENDAR 37 FOO 1234` or "fix" the day to display `MYCALENDAR 36 FOO 1234`?

Slightly off-topic.  An author once used my genealogy software to document the characters in a fantasy novel.  This included a custom calendar.  I don't know how many days were in the months of this calendar, but it may have been over 36.

fisharebest:
Perhaps systems should be instructed to treat unrecognised dates as if they were a `DATE.PHRASE`?

dthaler:
Merging per agreement in GEDCOM Steering Committee meeting 22 MAR 2022

PR #127, opened 2022-03-18
Initial strawman of EXID-TYPE registry

### LINKED ISSUE #121: Process for registering EXID.TYPE values
EXIDs are only interoperable without data loss if implementations use the same EXID.TYPE values to mean the same thing.
Currently there seems to be nowhere that such TYPE values can be listed.  How would one go about registering one for other implementations to use to mean the same thing?

Currently all that seems to exist is https://gedcom.io/migrate/#afn-rfn-rin which lists values for AFN, RIN, and RFN, but how would one add to that set in a non-implementation-specific way?

dthaler:
Please review list of fields, whether you like JSON as the format or something else, etc.
It might be nice if gedcom.io had some html page generated from the machine-readable registry file, but I didn't do that.

Fixes #121 

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1703,6 +1703,8 @@ the authority owning the identifier is provided in the TYPE substructure; see `E
 
 Depending on the maintaining authority, an `EXID` may be a unique identifier for the subject, an identifier for 1 of several views of the subject, or an identifier for the externally-maintained copy of the same information as is contained in this structure. However, unlike `UID` and `REFN`, `EXID` does not identify a structure; structures with the same `EXID` may have originated independently rather than by edits from the same starting point.
 
+`EXID` identifiers are expected to be unique and durable.
+
 #### `FAM` (Family record) `g7:record-FAM`
 
 See `FAMILY_RECORD`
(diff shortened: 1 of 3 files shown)


dthaler:
GEDCOM Steering Committee discussion 22 MAR 2022: no comments from those present other than looks ok.   @tychonievich please review.

tychonievich (approved):

specification/gedcom-03-datamodel.md
```diff
@@ -1703,6 +1703,8 @@ the authority owning the identifier is provided in the TYPE substructure; see `E
 
 Depending on the maintaining authority, an `EXID` may be a unique identifier for the subject, an identifier for 1 of several views of the subject, or an identifier for the externally-maintained copy of the same information as is contained in this structure. However, unlike `UID` and `REFN`, `EXID` does not identify a structure; structures with the same `EXID` may have originated independently rather than by edits from the same starting point.
 
+`EXID` identifiers are expected to be unique and durable.
```
tychonievich, line 1706:
```suggestion
`EXID` identifiers are expected to be unique. Once assigned, an `EXID` identifier should never be re-used for any other purpose.
```

Issue #136, opened 2022-05-03
Multiple surnames


fisharebest:
Spanish and Portuguese cultures have two surnames - one from each parent.

These are two independent surnames, not one "double-barrelled" surname, as is common in many English-speaking countries.

These two surnames may sometimes be separated by a particle/conjunction.  An example from [wikipedia](https://en.wikipedia.org/wiki/Spanish_naming_customs#Nominal_conjunctions):

`Óscar Arnulfo Romero y Galdámez`

> The character U+002F ( / , slash or solidus) has special meaning in a personal name, be‐
ing used to delimit the portion of the name that most closely matches the concept of a
surname, family name, or the like.

Since we have two surnames that are independent (and may not be consecutive) my application uses slashes around each of them.

```
1 NAME Óscar Arnulfo /Romero/ y /Galdámez/
2 GIVN Óscar Arnulfo
2 SURN Romero, Galdámez
```

Is the use of multiple slash-delimited surnames valid in GEDCOM 7.0?

If yes, I can create a PR to clarify the text - "portion of the name" => "portion (or portions) of the name".
If no, what would be the correct way to record such names

dthaler:
I believe the current answer is No, but the GEDCOM 7 spec does have problems here.
[Section 2.8](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#personal-name) answers the question where it specifies:

```
NamePersonal = nameStr
             / [nameStr] "/" [nameStr] "/" [nameStr]

nameChar     = %x20-2E / %x30-10FFFF  ; any but '/' and '\t'
nameStr      = 1*nameChar
```

That production explicitly only allows a single "/" delimited portion.

A bug, in my view, is in the spec in the production of [PERSONAL_NAME_STRUCTURE](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#PERSONAL_NAME_STRUCTURE) where it has:

```
n NAME <PersonalName>                      {1:1}  [g7:INDI-NAME](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#INDI-NAME)
  +1 TYPE <[Enum](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#cb17-1)>                           {0:1}  [g7:NAME-TYPE](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#NAME-TYPE)
[...]

PR #143, opened 2022-05-17
Revert to 0:1 cardinality for name pieces 

tychonievich:
As noted in #136, the {0:M} cardinality introduced an ambiguity. It was also omitted from the changelog when 7.0.0 was released. To the best of my knowledge it was not implemented by any application, and so is suitable for a patch release per the following version number clause

> If there is an ambiguity or contradiction in the specification, it will be resolved in a patch version unless it is known that implementations interpreted the spec differently and that clarifying the intended meaning would cause incompatibilities between those implementations.

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1050,7 +1050,7 @@ Names may have multiple parts that fit in the same piece; it is recommended that
 
 :::example
 ```gedcom
-1 NAME William "Bill" Edward Hernandez y Martinez
+1 NAME William "Bill" Edward /Hernandez y Martinez/
 2 GIVN William Edward
 2 NICK Bill
 2 SURN Hernandez Martinez
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1024,12 +1024,12 @@ A `NOTE_STRUCTURE` can contain a `SOURCE_CITATION`, which in turn can contain a 
 #### `PERSONAL_NAME_PIECES` :=
 
 ```gedstruct
-n NPFX <Text>                              {0:M}  g7:NPFX
-n GIVN <Text>                              {0:M}  g7:GIVN
-n NICK <Text>                              {0:M}  g7:NICK
-n SPFX <Text>                              {0:M}  g7:SPFX
-n SURN <Text>                              {0:M}  g7:SURN
-n NSFX <Text>                              {0:M}  g7:NSFX
+n NPFX <Text>                              {0:1}  g7:NPFX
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -1046,6 +1046,17 @@ Optional isolated name parts; see `PERSONAL_NAME_STRUCTURE` for more.
 ```
 :::
 
+Names may have multiple parts that fit in the same piece; it is recommended that they appear in the same order in a piece structure as they do in the piece's superstructure.
+
+:::example
+```gedcom
+1 NAME William "Bill" Edward Hernandez y Martinez
```
dthaler, line 1053:
```suggestion
1 NAME William "Bill" Edward /Hernandez y Martinez/
```

specification/gedcom-03-datamodel.md
```diff
@@ -1046,6 +1046,17 @@ Optional isolated name parts; see `PERSONAL_NAME_STRUCTURE` for more.
...
+2 GIVN William Edward
+2 NICK Bill
+2 SURN Hernandez Martinez
```
dthaler, line 1056: the above example requires explanation as to why it isn't "Hernandez y Martinez".  Otherwise it looks like a typo.

tychonievich: It's not clear to me if this is the right example. Another example (a historical figure of some prominence) might be

```gedcom
1 NAME Pablo Diego José Francisco de Paula Juan Nepomuceno Crispín Cipriano de la Santísima Trinidad María de los Remedios Alarcón y Herrera Ruiz /Picasso/
2 GIVN Pablo Diego José Francisco de Paula Juan Nepomuceno Cipriano de la Santísima Trinidad
2 SURN Ruiz Picasso
```

but I'm not 100% sure I did that right. "Cipriano" and "María de los Remedios Alarcón y Herrera" were names added at baptism, not birth, but might still count as "given" names, depending on how you think of "given". Additionally, while it is clear he went by Picaso and that Ruiz was another surname, is slashes around one but both in SURN correct?

specification/gedcom-03-datamodel.md
```diff
@@ -1046,6 +1046,17 @@ Optional isolated name parts; see `PERSONAL_NAME_STRUCTURE` for more.
 ```
 :::
 
+Names may have multiple parts that fit in the same piece; it is recommended that they appear in the same order in a piece structure as they do in the piece's superstructure.
+
+:::example
+```gedcom
+1 NAME William "Bill" Edward /Hernandez y Martinez/
```
tychonievich, line 1053:
```suggestion
1 NAME William "Bill" Edward /Hernandez Martinez/
```
Remove the conjunction

Issue #154, opened 2022-06-02
Problematic example in PERSONAL_NAME_PIECES


chronoplexsoftware:
It has previously been stated in [#134](https://github.com/FamilySearch/GEDCOM/issues/134#issuecomment-1116636979) that:

> We cannot restrict the use of quotes or add semantics to them in a patch or minor release. Any prescriptive change would have to wait for 8.0, and a larger name revision is being discussed for that version.
> 
> While quotes may be appropriate in some cultures, parentheses are more common in other cultures. We identified places where other delimiters or no delimiters were used; situations where nicknames were of varying importance to include in the name; situations where nicknames vary be relationship to the individual or that change over time; etc. We do no immediately see a simple solution to that.

7.0.8 adds the following example for `PERSONAL_NAME_PIECES`:
 `1 NAME William "Bill" Edward /Hernandez Martinez/` 

Given the above statement and example, which does enclose the nickname with `"`, a further remark is needed making it clear that a receiving system may not treat words enclosed in `"` as a nickname (unless of course there is a optional subordinate `NICK` tag with a payload matching the quoted value).

Without such a comment, readers may mistakenly believe that nicknames can be extracted from the `NAME` payload, especially since GEDCOM 5.4 and 5.5 did require the nickname to be enclosed in `"` when in the `NAME` payload.

tychonievich:
Resolved by #156

Issue #152, opened 2022-06-02
Name part cardinality change


chronoplexsoftware:
### Discussed in https://github.com/FamilySearch/GEDCOM/discussions/146

<div type='discussions-op-text'>

<sup>Originally posted by **tychonievich** May 24, 2022</sup>
The steering committee has made a change to correct an ambiguity in 7.0.0 through 7.0.7: we've changed the cardinality of `GIVN` and other name parts from `{0:M}` to `{0:1}`, as it was under 5.5.1 and earlier (#143). Previously the multiple cardinalities were not in the changelog for 7.0.0 and no semantics for them was defined.

We believe that this is permitted as a patch under the clause

> If there is an ambiguity or contradiction in the specification, it will be resolved in a patch version unless it is known that implementations interpreted the spec differently and that clarifying the intended meaning would cause incompatibilities between those implementations.

If you (know of) an application that implemented multiple copies of the same name part substructure in their 7.0 tool, please let us know.</div>

We consider that any change of cardinality from 'any number' to 'singular' should always be categorised as a breaking change, reserved for major releases, since it removes a valid use of a tag.  Files containing multiple instances of name parts would be valid in 7.0.0-7.0.7 but not in 7.0.8 or 7.x, therefore a patch or minor version is not an appropriate mechanism for making this change.  We would request that the change of cardinality be reverted to preserve the integrity of the semanitc versioning added for GEDCOM 7.

Further, there is at least one implementation of 7.0 which has made use of this feature.  The approach is documented here: [https://chronoplexsoftware.com/gedcomextensions/](https://chronoplexsoftware.com/gedcomextensions/)

In this implementation, the `NAME` payload still contains the correctly formatted and primary representation of the name but multiple instances of `GIVN` or `SURN` and an extension `_NAME_PART_TYPE` are used to distinguish between first names, middle names, and rufname / callname, and patronymic / matronymic surname parts (and indeed any future name part that a client wishes to represent).  This is because, unfortunately, GEDCOM 7.0 did not add a standard way to support this.  This is entirely valid use of `{0:M}` cardinality in 7.0.0-7.0.7, irrespective of whether the original cardiality change was documented.

chronoplexsoftware:
[...]

PR #156, opened 2022-06-03
Revert name part cardinality

tychonievich:
Per #152, the belief that we could change name cardinalities in a patch was based on a misapprehension of their use in existing applications. This PR proposes reverting that with a new 7.0.9 release.

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1024,12 +1024,12 @@ A `NOTE_STRUCTURE` can contain a `SOURCE_CITATION`, which in turn can contain a 
 #### `PERSONAL_NAME_PIECES` :=
 
 ```gedstruct
-n NPFX <Text>                              {0:1}  g7:NPFX
-n GIVN <Text>                              {0:1}  g7:GIVN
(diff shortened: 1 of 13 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -1046,16 +1046,9 @@ Optional isolated name parts; see `PERSONAL_NAME_STRUCTURE` for more.
...
-:::example
-```gedcom
-1 NAME William "Bill" Edward /Hernandez Martinez/
-2 GIVN William Edward
-2 NICK Bill
-2 SURN Hernandez Martinez
-```
-:::
+This specification does not define how the meaning of multiple parts with the same tag differs from the meaning of a single part with a concatenated larger payload.
+However, some applications allow the user to chose whether to combine or split name parts, meaning the tag quantity it should be treated as expressing at least a user preference.
```
dthaler, line 1050:
```suggestion
However, some applications allow the user to chose whether to combine or split name parts, meaning the tag quantity should be treated as expressing at least a user preference.
```

dthaler (approved):

tychonievich:
Note this also resolves #154

PR #162, opened 2022-06-19
Recommend SDATE use single Gregorian dates

### LINKED ISSUE #160: Recommend SDATE use single Gregorian dates
### Discussed in https://github.com/FamilySearch/GEDCOM/discussions/140

Those who have engaged in a discussion about `SDATE` generally agree that sorting is easiest if all sort-by dates are single dates (not ranges or periods) in the same calendar. We cannot change this in a patch or minor release as it is not backwards compatible and we know of applications that allow users to enter any kind of date in a sort-by field, but we can make a recommendation like

> It is recommended that the `SDATE`'s payload be in the Gregorian calendar and be a simple `date` with the `calendar` omitted. Including a `month` and `day` is encouraged to help different applications sort dates the same way, as the relative ordering of dates with different levels of precision is not well defined.

dthaler:
Fixes #160

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -2420,6 +2420,11 @@ while the `DATE` should be displayed as the date of the structure.
 
 `SDATE` and its substructures (including `PHRASE`, `TIME`, and any extension structures) should be used only as sorting hints, not to convey historical meaning.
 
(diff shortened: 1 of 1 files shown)


jamestanner45 (approved):

dthaler:
Merging per agreement in meeting 21 JUN 2022

PR #171, opened 2022-07-13
Clarify documented extension tag URIs

tychonievich:
A private conversation with some implementers suggested that we needed to clarify that

1. Each structure type and enumeration value requires a distinct URI: we can't simply use the URI of an entire extension vocabulary for all of its tags.
2. Each URI needn't be the URL of a separate document the way the standard tags do (e.g. <https://gedcom.io/terms/v7/record-INDI>); rather, many URIs can lead to the same document.

The first of these points was only indirectly present in the previous text; this PR makes it explicit.
The second point was absent entirely, and is not strictly part of the specification at all; this PR adds a non-normative informative link to a W3C interest group note on the subject with a brief summary of what a reader might find there.

### DIFF
--- a/specification/gedcom-01-serialization.md
+++ b/specification/gedcom-01-serialization.md
@@ -445,7 +445,7 @@ or what they call "Hash URIs" (which require no setup).
 
 That interest group note also explains why it might be desirable
 to have a separate URIs for a concept and the document describing that concept.
-Because of the structure of the schema, that separation is less important for FamilySearch GEDOCM 7
+Because of the structure of the schema, that separation is less important for FamilySearch GEDCOM 7
 than it is for the semantic web, but it remains good advice where feasible.
 :::
 
--- a/specification/gedcom-01-serialization.md
+++ b/specification/gedcom-01-serialization.md
@@ -401,12 +401,17 @@ defines the following tags
 | :---- | :---- |
 | `_SKYPEID` | `http://xmlns.com/foaf/0.1/skypeID` |
 | `_MEMBER` | `http://xmlns.com/foaf/0.1/member` |
+
+Note that at the time of writing, the [FOAF](https://xmlns.com/foaf/spec/20140114.html) URIs used in this example are not URLs.
 :::
 
+
 The meaning of a documented extension tag is identified by its URI, not its tag.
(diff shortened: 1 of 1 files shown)


specification/gedcom-01-serialization.md
```diff
@@ -401,12 +401,17 @@ defines the following tags
 | :---- | :---- |
 | `_SKYPEID` | `http://xmlns.com/foaf/0.1/skypeID` |
 | `_MEMBER` | `http://xmlns.com/foaf/0.1/member` |
+
+Note that at the time of writing, the [FOAF](https://xmlns.com/foaf/spec/20140114.html) URIs used in this example are not URLs.
```
tychonievich, line 405: This text might not be the best.

- We could expand on it, explaining that URI meanings survive changes such as the reconfiguring of the xmlns.com host in 2022 that made these URIs no longer redirect to the FOAF spec.
- We could replace the FOAF examples with another semantic web ontoloy like schema.org that hasn't (yet) had [link rot](https://en.wikipedia.org/wiki/Link_rot).
- We could use `example.com` URIs instead. That might be preferable because while FOAF does define skypeID, it does so without reference to GEDCOM and hence without indicating whether it's a structure type or enumerated value, where it can appear, etc.
- We could wait to update this until a 7.0 extension is published so we can use a real GEDCOM example

I don't have strong feelings on which (if any) of the above options is better than the PR's current text

specification/gedcom-01-serialization.md
```diff
@@ -429,6 +434,21 @@ and a system importing one may export it as the other without change of meaning.
...
+:::note
+The W3C has an [interest group note](http://www.w3.org/TR/cooluris/)
+that discusses several ways of achieving this URI/URL mapping,
+including how a single webpage can describe multiple tags
+using either HTTP redirects (which requires some server setup)
+or what they call "Hash URIs" (which require no setup).
+
+That interest group note also explains why it might be desirable
+to have a separate URIs for a concept and the document describing that concept.
+Because of the structure of the schema, that separation is less important for FamilySearch GEDOCM 7
```
dthaler, line 448:
```suggestion
Because of the structure of the schema, that separation is less important for FamilySearch GEDCOM 7
```

dthaler (approved):
Ok except for a typo that needs to be fixed.

Issue #172, opened 2022-07-13
Sorting in INDIVIDUAL_ATTRIBUTE_STRUCTURE and INDIVIDUAL_EVENT_STRUCTURE


ungeahnt:
One question: Is there a specific reason for the sort order in [INDIVIDUAL_ATTRIBUTE_STRUCTURE](https://github.com/FamilySearch/GEDCOM/blob/main/specification/gedcom-03-datamodel.md#individual_attribute_structure-)? It looks like some tags are sorted alphabetically and others are not (e.g. ADOP, BIRT, CHR, EVEN are listed at the very end). Is this intentional?

Same with FACT at [INDIVIDUAL_EVENT_STRUCTURE](https://github.com/FamilySearch/GEDCOM/blob/main/specification/gedcom-03-datamodel.md#individual_event_structure-)

ungeahnt:
My point is about the readability of the GEDCOM specification. 

I had looked for `BIRT `in `INDIVIDUAL_ATTRIBUTE_STRUCTURE `and at first glance found nothing, thinking that the tags in this list were sorted alphabetically. That BIRT appeared at the very end was a bit irritating. Also, I couldn't figure out a rule for this type of sorting. 

A consistent sorting per list/structure incl. a comment in the text would certainly be helpful.

albertemmerich:
It will not be easier if we now modify the ordering of events and/or facts in the spec. We will have a lot of modifications in the document, and these have to be checked whether there is any need to adjust applications which have already implemented these structures. This may be better for those who read the document for the first time, but for all others it makes work more complicated.

PR #174, opened 2022-07-15
Massive reordering of metasyntax blocks

tychonievich:
Issue #172 recommended alphabetizing structures within metasyntax blocks. Doing that often broke up semantically-related structures, so I tried to document excepsions. The end result was a seven-part ordering heuristic (documented in the changed text itself) and I'm still not entirely happy with all of how it turned out. There was something familiar with `NOTE_STRUCTURE` always being near the end and so on...

However, the clarifying text that order doesn't matter except for the `Dataset :=` rule should definitely stay. And the reordering of alternatives in a choice production was a lot less tricky than the reordering of substructures, so that should probably happen even if the rest of this is rejected.

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -92,7 +92,7 @@ The intent of this metasyntax is to resemble the line encoding of allowable stru
         - `ADDRESS_STRUCTURE`'s substructures are presented in the same order they would be serialized to create a postal address
         - `NPFX` is listed before other name parts; `SPFX` before `SURN`, and `NSFX` after other name parts
     
-    - Some structures are placed together because of sematic similarity. For example,
+    - Some structures are placed together because of semantic similarity. For example,
         - `LANG` is with `TRAN`
         - `CHAN` is with `CREA`
         - `ASNI` is with `DESI`
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -98,7 +98,7 @@ The intent of this metasyntax is to resemble the line encoding of allowable stru
         - `ASNI` is with `DESI`
         - `HUSB` is with `WIFE`
         - `ADDRESS_STRUCTURE` is with `EMAIL`, `FAX`, `PHON`, and `WWW`
-        - `DATE` with `SDATE`
+        - `DATE` is with `SDATE`
     
     - Meaning modifiers are placed before other substructures. For example,
         - `TIME`, `ROLE`, `TYPE`, `FORM`, and `PHRASE` are placed before other substructures.
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -85,7 +85,7 @@ The intent of this metasyntax is to resemble the line encoding of allowable stru
 
     Presentation order is chosen as follows:
     
-    - If there is text suggesting an order, that order is followed in the metasyntax. For example,
+    - If there is text suggesting an order among structures, that order is followed in the metasyntax. For example,
         - `HEAD.GEDC` is presented as the first substructure of `HEAD`.
     
     - If a structure represents something with a traditional order, that order is used. For example,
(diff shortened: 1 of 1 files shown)


ungeahnt:
> Issue https://github.com/FamilySearch/GEDCOM/issues/172 recommended alphabetizing structures within metasyntax blocks

Only for clarification: I was not talking about a **single universal** (alphabetical or otherwise) sort for all structures. I think this is also not meaningful.

For  `INDIVIDUAL_ATTRIBUTE_STRUCTURE`, the majority was sorted alphabetically, so at first glance I *assumed* an alphabetical ordering. But this does not mean that `INDIVIDUAL_ATTRIBUTE_STRUCTURE` has to be sorted alphabetically in the future and certainly not other lists/structures.

As I said:

>A consistent sorting **_per_** list/structure incl. a comment in the text would certainly be helpful.

That means per list/structure the kind of sorting can vary.

specification/gedcom-03-datamodel.md
```diff
@@ -81,6 +81,39 @@ The intent of this metasyntax is to resemble the line encoding of allowable stru
         
         Pseudo-structures do not have a URI.
 
+- Within the metasyntax, the order in which substructures are presented within a structure and the order in which choices are presented within an option set are not significant unless otherwise specified in the text next to the metasyntax block.
+
+    Presentation order is chosen as follows:
+    
+    - If there is text suggesting an order, that order is followed in the metasyntax. For example,
```
dthaler, line 88:
```suggestion
    - If there is text suggesting an order among structures, that order is followed in the metasyntax. For example,
```

specification/gedcom-03-datamodel.md
```diff
@@ -81,6 +81,39 @@ The intent of this metasyntax is to resemble the line encoding of allowable stru
...
+        - `ADDRESS_STRUCTURE`'s substructures are presented in the same order they would be serialized to create a postal address
+        - `NPFX` is listed before other name parts; `SPFX` before `SURN`, and `NSFX` after other name parts
+    
+    - Some structures are placed together because of sematic similarity. For example,
+        - `LANG` is with `TRAN`
+        - `CHAN` is with `CREA`
+        - `ASNI` is with `DESI`
+        - `HUSB` is with `WIFE`
+        - `ADDRESS_STRUCTURE` is with `EMAIL`, `FAX`, `PHON`, and `WWW`
+        - `DATE` with `SDATE`
```
dthaler, line 101:
```suggestion
        - `DATE` is with `SDATE`
```

specification/gedcom-03-datamodel.md
```diff
@@ -141,26 +186,19 @@ n HEAD                                     {1:1}
...
-  +1 PLAC                                  {0:1}  g7:HEAD-PLAC
-     +2 FORM <List:Text>                   {1:1}  g7:HEAD-PLAC-FORM
-  +1 <<NOTE_STRUCTURE>>                    {0:1}
 ```
 
 The header pseudo-structure provides metadata about the entire dataset.
 A few substructures of note:
 
 - `GEDC` identifies the specification that this document conforms to.
+    It is recommended that `GEDC` be the first substructure of the header.
```
dthaler, line 196: If this is only a recommendation, it may be worth explaining this one, since the GEDC.VERS is used to determine how to parse the rest of the file, anything before it may not be parsed "correctly" per the version detection algorithm. Personally, I think it should be a _requirement_ in FamilySearch GEDCOM 7 that it be first.   It was already listed first so I might argue it was an ambiguity/clarification, not a breaking change.

dthaler (commented):
I like the direction this is going.  I just want to discuss my GEDC.VERS comment before approving.

specification/gedcom-03-datamodel.md
tychonievich: The recommendation was already in the spec; I just copied it up here from <https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#GEDC>. I do think requiring it to be first would be a breaking change.

tychonievich:
@ungeahnt, thanks for the [clarification](#issuecomment-1185683608) on what you meant. I was not trying to imply all of these changes were in response to your request, only that your issue was impetus that got me thinking along these lines.

tychonievich:
Discussed and rejected in steering committee meeting.

We should note that order is not significant, and alphabetize the mostly-alphabetical lists of events and attributes, but no other changes. We will make a separate pull request for those (much smaller) changes.

PR #179, opened 2022-07-19
Sort events and note order is not significant

tychonievich:
See #172 for the suggestion to do this and #174 for a more involved sorting we considered but rejected.


albertemmerich (approved):
Fine for me.
Albert

dthaler (approved):

Issue #183, opened 2022-07-21
Documenting, correcting, and using YAML


tychonievich:
When I was preparing the 7.0.0 spec for publication, I created an ad-hoc YAML format to host at the various URIs listed in the spec. Since then we've [added hand-crafted YAML files](https://github.com/FamilySearch/GEDCOM.io/pull/43) for some EXID types and most recently [Albert proposed](https://github.com/FamilySearch/GEDCOM/issues/175#issuecomment-1191338286) using YAML for an extension tag registry.

Before we use these YAML documents further, we should fix some problems in their design and document their structure.

In the current YAML, some things don't seem right

1. Inextensibility: `payload:` has three options; the third uses a tag so is not usable with documented extensions
    - a type URI
    - `Y|<NULL>`
    - `@<XREF:`**tag**`>@`
    
    Proposal: change `@<XREF:`**tag**`>@` to `@`**uri**`@`

2. Inextensibility: enumerations do not list the structure(s) they may be used in, though the structures do list the enumeration values they allow. This asymmetry means an extension can make a new enumeration tag, but cannot add it to the set allowed by a standard structure.

    Proposal: add a mapping key `structures` with a list of URIs of structure types that allow this enumerated value

3. Poor design: Months use `type: enumeration` even though they are not enumerations.

4. Poor design: EXID TYPE payloads use `type: legacy` instead of something meaningful like `known-uri` or `exid-type`.

5. Omission: Months use do not list the calendar(s) they apply to

    Proposal: add a mapping key `calendars` with a list of URIs of calendars that include this month.
    
6. Omission: Calendars have the list of months in their description, but not in machine-readable fields

    Proposal: add a mapping key `months` with a list of month URIs used in this calendar. Specify that this list is in month order within the year to facilitate date sorting.
    
If we want to use YAML for the extension registry too, we should

7. Add "extension tags" as a mapping key in `type: structure`, and specify that it is an ordered list of known tags and the first should be used when possible.

8. Document the types used and the YAML mapping keys expected in each, including some of the non-obvious things like how we denote records and substructures of the header.

9. Clarify that because extensions are added over time, the superstructure/substructure and enumeration/structure references may be asymmetric: if either is present, the extension allows the relationship.

10. Clarify that it is not permitted to
    1. add a new extension month to an existing calendar: calendar-month relationships must be symmetric in the YAML.
    2. have a superstructure/substructure relationship specified with different in cardinalities on each side: cardinality cannot be changed as an extension.

[...]

PR #184, opened 2022-07-21
Fix YAML formatting, missing information

tychonievich:
Handles items 2, 3, 5, 6, 11, and 12 from #183. Those were chosen because they either fix YAML formatting errors or add new data without modifying existing data, and hence should be non-breaking for any application using the YAML files.

One potentially breaking change in this PR is that months now use `type: month` instead of `type: enumeration`, which could impact some YAML-consuming application.


tychonievich:

1. should separate out the YAML description file, move it to a separate PR
2. should merge the YAML fixes
3. should change `@<XREF:tag>@` to `@<uri>@`
4. should host translations in separate files (probably, maybe revisit in re user-centric explanations etc)

Issue #180, opened 2022-07-20
Permit undocumented extension tags the same flexibility as standard tags


tychonievich:
When I was trying to formally define structure types during the early stages of 7.0 drafting, I realized that at least back to GEDCOM 5.3 there were implicit rules about tag meanings:

- A tag can have multiple meanings, including multiple enumerated and multiple structure type meanings.
- Which meaning is meant is uniquely identified by (a) the tag and (b) the structure type containing it.

See [#178](https://github.com/FamilySearch/GEDCOM/issues/178#issuecomment-1190293633) for additional comments on what "meaning" means.

**Question:** can undocumented extension tags have multiple meanings like this?

**As written** I think the answer is No, they cannot. About undocumented extension tags the spec says say "The meaning of an undocumented extension tag is identified by its tag" without adding any "and the structure type containing it".

**As implemented** I think the answer is Yes, they can. One well-documented  and widely implemented example is [GEDCOM-L](https://genealogy.net/GEDCOM/)'s `_LOC` tag, used for both a record type and a substructure type that points to the record type.

This seems like an **error** in the spec to me: it is not particularly ambiguous, but it also does not describe current practice. I propose that we should loosen the spec as a patch to remove this error and allow undocumented extensions to continue to have this context-specific meaning.

tychonievich:
Discussed in steering committee meeting
Decided to add a patch to the spec to allow the current common use: tags meaning can vary by the structure type containing it, even for extension tags.

PR #188, opened 2022-08-23
Extension structure meaning depends on superstructure type

tychonievich:
As discussed in issue #180, practice is that extension tags like `_LOC` means different things under different superstructures, but the current text does not acknowledge that. This PR adds that acknowledgement.

### DIFF
--- a/specification/gedcom-01-serialization.md
+++ b/specification/gedcom-01-serialization.md
@@ -430,7 +430,7 @@ and a system importing one may export it as the other without change of meaning.
 :::
 
 An extension tag that is not given a URI in the schema structure is called an **undocumented extension tag**.
-The meaning of an undocumented extension tag is identified by its superstructure type and tag.
+The meaning of an undocumented extension tag is identified by its superstructure type and its tag.
 
 
 ### Requirements and Recommendations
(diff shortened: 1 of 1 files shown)


specification/gedcom-01-serialization.md
```diff
@@ -403,7 +403,7 @@ defines the following tags
 | `_MEMBER` | `http://xmlns.com/foaf/0.1/member` |
 :::
 
-The meaning of a documented extension tag is identified by its URI, not its tag.
+The meaning of a documented extension tag is identified by its superstructure type and URI, not its tag.
```
dthaler, line 406:
```suggestion
The meaning of a documented extension tag is identified by its superstructure type and its URI, not its tag.
```
to avoid misreading it as the URI of the superstructure

specification/gedcom-01-serialization.md
```diff
@@ -430,7 +430,7 @@ and a system importing one may export it as the other without change of meaning.
 :::
 
 An extension tag that is not given a URI in the schema structure is called an **undocumented extension tag**.
-The meaning of an undocumented extension tag is identified by its tag.
+The meaning of an undocumented extension tag is identified by its superstructure type and tag.
```
dthaler, line 433:
```suggestion
The meaning of an undocumented extension tag is identified by its superstructure type and its tag.
```

lynchrs3 (approved):

Issue #206, opened 2022-09-25
Empty strings in ABNF


DavidMStraub:
Hi,

sorry in advance if this is a naive question. I've started looking into parsing GEDCOM 7 with Python and had some issues reading the `grammar.abnf` with the Python [abnf](https://github.com/declaresub/abnf) package. I traced it back to the presence of the empty string `""` as option in three cases.

- for `DateValue` and `DatePeriod`, `""` is given as last option. Could this be wrapped into `[` `]` as optional instead of having the empty string as option?
- For `listItem`, `""` is the first option. What is the motivation for that?

tychonievich:
Empty strings as `""` instead of enclosing `[`...`]` were employed to make it clear to the reader that the empty string was a valid payload for those types. That's also why we included `""` as an option for DateValue even though it was also permitted for `DatePeriod`, one of `DateValue`'s options.

[RFC 5234](https://www.rfc-editor.org/rfc/rfc5234) does allow empty quoted strings, as given by the `*` instead of `1*` in the defining rule for `char-val`:

```abnf
char-val       =  DQUOTE *(%x20-21 / %x23-7E) DQUOTE
```

That said, in principle I see no objection to using optional notation instead, particularly if common tools fail to implement this part of ABNF: we can draw attention to the potential for an empty payload using a comment instead.

The `List:Text` payload type is used for `PLAC` and `PLAC.FORM`. `PLAC` is sometimes used with an empty payload supply its `MAP` and/or `EXID` substructures without providing a place name. We could argue about the necessity of that use (e.g. an empty `PLAC` payload could always be replaced by "`PLAC Earth`") but banning it now would violate semantic versioning.

DavidMStraub:
Thanks @tychonievich!

PR #207, opened 2022-09-26
Use [ ] instead of "" to denote optional payloads

tychonievich:
This was requested in #206 to support ABNF packages that fail to handle `""` correctly. It changes the form of the ABNF, but not its meaning.

### DIFF
--- a/extracted-files/grammar.abnf
+++ b/extracted-files/grammar.abnf
@@ -59,11 +59,11 @@ Enum    = Tag
 
 ; ------------- Date -------------
 
-DateValue   = date / DatePeriod / dateRange / dateApprox / ""
+DateValue   = [ date ] / DatePeriod / dateRange / dateApprox
 DateExact   = day D month D year  ; in Gregorian calendar
-DatePeriod  = %s"FROM" D date [D %s"TO" D date]
-            / %s"TO" D date
-            / ""
+DatePeriod  = [ %s"TO" D date ]
+            / %s"FROM" D date [D %s"TO" D date]
+            ; note both DateValue and DatePeriod can be the empty string
 
 date        = [calendar D] [[day D] month D] year [D epoch]
 dateRange   = %s"BET" D date D %s"AND" D date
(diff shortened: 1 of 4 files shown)


extracted-files/grammar.abnf
```diff
@@ -59,11 +59,11 @@ Enum    = Tag
 
 ; ------------- Date -------------
 
-DateValue   = date / DatePeriod / dateRange / dateApprox / ""
+DateValue   = [ date ] / DatePeriod / dateRange / dateApprox
```
dthaler, line 62: I think
```suggestion
DateValue   = [ date / DatePeriod / dateRange / dateApprox ]
```
might be clearer

extracted-files/grammar.abnf
```diff
@@ -59,11 +59,11 @@ Enum    = Tag
...
 DateExact   = day D month D year  ; in Gregorian calendar
-DatePeriod  = %s"FROM" D date [D %s"TO" D date]
-            / %s"TO" D date
-            / ""
+DatePeriod  = [ %s"TO" D date ]
+            / %s"FROM" D date [D %s"TO" D date]
+            ; note both DateValue and DatePeriod can be the empty string
```
dthaler, line 66: would this be the same as the simpler form
```suggestion
DatePeriod  = [%s"FROM" D date] [D %s"TO" D date]
```
?

extracted-files/grammar.abnf
```diff
@@ -111,13 +111,13 @@ days    = Integer %x64    ; 21d
 
 ; ------------- List -------------
 
-List      = listItem *(listDelim listItem)
-listItem  = "" / nocommasp / nocommasp *nocomma nocommasp
+list      = listItem *(listDelim listItem)
+listItem  = [ nocommasp ] / nocommasp *nocomma nocommasp ; may be empty
```
dthaler, line 115:
```suggestion
listItem  = [ nocommasp / nocommasp *nocomma nocommasp ]
```

extracted-files/grammar.abnf
tychonievich: That's simpler, but not the same; the former encodes "`TO 2022`" but the latter " TO 2022`" (note the extra space)

dthaler (approved):

Issue #221, opened 2022-10-11
Add RESN to SOUR, REPO, SNOTE and SUBM


fisharebest:
The `RESN` tag can only be added to `FAM`, `INDI` and `OBJE` records.

I would like to propose that it also be added to the other record types: `SOUR`, `REPO`, `SNOTE` and `SUBM`.

Sources, repositories, etc. can all be public/private.  For example, my grandmother's diary can be found in my personal files, both of which are private.

----------
/disclaimer - this is an extension that I already use in my own software ;-)

tychonievich:
SOUR.RESN and REPO.RESN were already proposed in #158 and were queued for 7.1. SNOTE.RESN and SUBM.RESN also make sense to me. And as per #64, if we add it to SNOTE we should also add it to NOTE

Issue #158, opened 2022-06-05
Allow RESN under other record types


dthaler:
`SUBMITTER_RECORD` and `REPOSITORY_RECORD` can contain name and contact information (phone number, address, etc.).  Personal contact information may be private, such as an unlisted telephone number, and might be revealed in a submitter record or a repository record (for privately held sources such as a Family Bible, the person who holds the source may be entered as a repository). 

The `RESN` structure was defined to mark records as confidential or private, but in 7.0, `RESN` can only appear under `FAMILY_RECORD`, `INDIVIDUAL_RECORD`, or `MULTIMEDIA_RECORD`.

It may also be worth adding under `SOURCE_RECORD` since some GEDCOM files may put personal contact information in a source record, such as a private email address of someone else when citing email, or a private phone number when citing a phone interview.

Norwegian-Sardines:
Since a Source_Record can also have TEXT associated with the record (defined as: "A verbatim copy of any description contained within the source.") and that text can include information that should be kept confidential, therefore instances of this record should also have RESN.  

The same could also be true of a information in the Note_Structure both "inline" and "shared" notes, where the data entry person could have notes to themselves that indicate something that should not be transferred our displayed to unauthorized individuals. This concept is not covered by RESN at the fact level since it is the NOTE not the Date/Place information that could be confidential!

PR #223, opened 2022-10-11
Add RESN to more structures

tychonievich:
Proposal to allow RESN in all records, and in `NOTE`s
See issues #158 and #221


dthaler (approved):

tychonievich:
Discussed 2022-10-25
We should add `RESN` to `g7:SOUR` as well as `g7:record-SOUR` because sometimes the citations (not just the records) contain sensitive information.

We also discussed that a `ADOP.FAMC` pointing to a restricted `FAM` will become `@VOID@`, but as written a `PHRASE` under it might be preserved anyway. Perhaps we should point this out in the spec?

specification/gedcom-03-datamodel.md
```diff
@@ -2347,6 +2353,11 @@ The `RESN` structure is provided to assist software in filtering data that shoul
 such that certain `RESN` structure payload entries result in the `RESN` structure and its superstructure being removed from the export.
 Such removal must abide by some constraints: see [Removing data](#removing-data) for more.
 
+A structure that points to a record that is being removed due to `RESN` processing will have its pointer replaced by a `voidPtr`,
+presering the structure's substructures.
+That might be desireable, but could cause accidental information leakage; for example, if a `g7:ADOP-FAMC` points to a restricted `FAMC` it may contain `FAMC`-identifying information in its substructures that the user also wishes to treat as restricted.
```
tychonievich, line 2358:
```suggestion
preserving the structure's substructures.
That might be desirable, but could cause accidental information leakage; for example, if a `g7:ADOP-FAMC` points to a restricted `FAMC` it may contain `FAMC`-identifying information in its substructures that the user also wishes to treat as restricted.
```

Issue #195, opened 2022-08-30
Inconsistency with ENUMs, PHRASEs and OTHER


fisharebest:
Some ENUM values include the value `OTHER`, and also allow a `PHRASE` subtag.

However, three elements allow a `PHRASE` subtag, but **do not** allow a value of `OTHER`.

For consistency, I think we should allow `OTHER` as the alternative would be to leave the element empty.

* `g7:FAMC-STAT`.  Here, it seems reasonable to add `OTHER` to the list of values.

* `g7:FAMC-ADOP`. Given the current values, `HUSB`, `WIFE`, `BOTH`, I cannot think of what other value might be needed.   So either add `OTHER` or remove `PHRASE`.

* `g7:SOUR-EVEN`. Given the current definition "the type of event or attribute", I cannot think of what other value might be needed.   So either add `OTHER` or remove `PHRASE`.

tychonievich:
Enumeration values besides `OTHER` can have a `PHRASE`; an example of this is present in the spec

```gedcom
2 ROLE FRIEND
3 PHRASE best friend
```

Hence, while `OTHER` needs a `PHRASE`, `PHRASE` may also be included where there is no `OTHER`. 

That said, it may be worth revisiting the placement of `OTHER` and `PHRASE`. I no longer recall the details of why we made the decisions we did, but can see 

- `g7:FAMC-ADOP`.`PHRASE` with pre-interpreted references, such as `BOTH` with `PHRASE by the Jones family` or `WIFE` with `PHRASE Mrs. Eli Whitkower`

- `g7:SOUR-EVEN`.`PHRASE` with clarity beyond event type, like `PHRASE clerk's copy, not original record` or `PHRASE primarily documents officator's activities; the couples being married were secondary`

I'm unsure about adding `OTHER` to `g7:FAMC-STAT`. I see the value in adding `PHRASE`s wthout picking one of the three current values. But as-is, `g7:FAMC-STAT` provides something of an ordering of confidence (`PROVEN` > no-`STAT` > `CHALLENGED` > `DISPROVEN`) and can be used to influence display, for example by hiding low-confidence links at a user-selected confidence level. If we add `OTHER` it doesn't seem to belong in that order in any obvious way. If we do want to add another value to allow phrases without using one of the current values, I'm tempted to pick one that fits in the hierarchy: maybe something like `UNCHALLENGED` or `DEFAULT` with an explicit statement that this is the implied meaning of a missing `STAT` too.

[...]

PR #224, opened 2022-10-11
Add FAMC.STAT UNCHALLENGED

tychonievich:
As discussed in #195

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -2971,7 +2971,7 @@ are not currently given their own tag. Cultural or personal gender preference ma
 | `CHALLENGED` | Linking this child to this family is suspect, but the linkage has been neither proven nor disproven. |
 | `DISPROVEN` | There has been a claim by some that this child belongs to this family, but the linkage has been disproven. |
 | `PROVEN` | There has been a claim by some that this child does not belong to this family, but the linkage has been proven. |
-| `UNCHALLENGED` | There have been no known claims that the link between this child and this family is suspect. `UNCHALLENGED` is implied if this structure is not present. |
+| `UNCHALLENGED` | There have been no known claims that the link between this child and this family is suspect. |
 
 :::note
 The structures for representing the strength of and confidence in various claims are known to be inadequate and are likely to change in a future version of this specification.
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -2971,6 +2971,7 @@ are not currently given their own tag. Cultural or personal gender preference ma
 | `CHALLENGED` | Linking this child to this family is suspect, but the linkage has been neither proven nor disproven. |
 | `DISPROVEN` | There has been a claim by some that this child belongs to this family, but the linkage has been disproven. |
 | `PROVEN` | There has been a claim by some that this child does not belong to this family, but the linkage has been proven. |
+| `UNCHALLENGED` | No reason to suspect the link between this child and this family has been identified. `UNCHALLENGED` is implied if this structure is not present. |
```
dthaler, line 2974: no reason to suspect it's been identified sounds like it's in doubt, is that what is meant?  I might suggest

```suggestion
| `UNCHALLENGED` | There have been no known claims that the link between this child and this family is suspect. `UNCHALLENGED` is implied if this structure is not present. |
```

I would also ask about the stated meaning of PROVEN.   If the linkage has been proven, but there has _not_ been a claim that this child does not belong to this family, what status should be used?

tychonievich: `PROVEN` was added with this wording in 5.5.1; not sure what the rationale was

dthaler: Tracking via issue #225 and proposed PR #228

dthaler (approved):

tychonievich:
Discussed 2022-10-18
"Unchallenged" and "unknown" are not the same idea, so this is not the correct solution yet. We might be able to come up with some kind of "none of the above" value (perhaps `OTHER`), or might need to add two values.

specification/gedcom-03-datamodel.md
```diff
@@ -2971,6 +2971,7 @@ are not currently given their own tag. Cultural or personal gender preference ma
 | `CHALLENGED` | Linking this child to this family is suspect, but the linkage has been neither proven nor disproven. |
 | `DISPROVEN` | There has been a claim by some that this child belongs to this family, but the linkage has been disproven. |
 | `PROVEN` | There has been a claim by some that this child does not belong to this family, but the linkage has been proven. |
+| `UNCHALLENGED` | There have been no known claims that the link between this child and this family is suspect. `UNCHALLENGED` is implied if this structure is not present. |
```
tychonievich, line 2974:
```suggestion
| `UNCHALLENGED` | There have been no known claims that the link between this child and this family is suspect. |
```

Remove the "if not present" clause due to comments in pull request conversation thread

dthaler:
@tychonievich needs rebase to resolve conflict

tychonievich:
See #195 for why this is being closed

Issue #229, opened 2022-10-13
QUAY enum values are not tags


DavidMStraub:
Hi,

while working on the [Python parser](https://github.com/DavidMStraub/python-gedcom7), I ran into a problem which I believe is an inconsistency in the current definition (or just a misunderstanding of mine).

The values of the `QUAY` enum (e.g. in the example file maximal70.ged) are integers. However the grammar defines `Enum = Tag` and `Tag` must start with an uppercase letter.

dthaler:
I agree this is a bug in the 7.0 spec that did not exist in the 5.5.1 spec.
I think the correct fix is to change Enum to be
`Enum = Tag / Integer`
or even just minimally
`Enum = Tag / digit`

fisharebest:
Might other/future Enums also want to be digits or integers?
This would rule out (3).

tychonievich:
I like Option 1. I think it's the least disruptive fix.

dthaler:
I would prefer option 1 more than allowing things like "2ND".   There's no demand for the permissiveness of "2ND" and it's not as good, in my view, as "2" or "SECOND", either of which are fine in option 1.   Since we can add permissiveness later, but can never remove permissiveness without a major version change, I'd argue for a less permissive fix now.

PR #233, opened 2022-10-18
Permit enums to begin with a digit

tychonievich:
As noted in #229, `QUAY` uses digit-beginning enumeration values so the spec should document that they are allowed.

This is the most permissive of the fixes discussed in #229; more restrictive changes like

> with the additional permission that standard enumerations may be single digits.
>
> ```abnf
> stdEnum = stdTag / digit
> ```

or 

> with the additional permission that standard enumerations may be non-negative integers.
>
> ```abnf
> stdEnum = stdTag / Integer
> ```

 might be better to minimize disruption.

### DIFF
--- a/specification/gedcom-02-datatypes.md
+++ b/specification/gedcom-02-datatypes.md
@@ -34,7 +34,7 @@ The URI for the `Integer` data type is `xsd:nonNegativeInteger`.
 
 An enumeration is a selection from a set of options.
 They are represented as a string matching the same production as a tag,
-with the additional permission that standard enumerations may begin with a digit.
+with the additional permission that standard enumerations may be integers.
 
 ```abnf
 stdEnum = stdTag / digit *tagchar
--- a/specification/gedcom-02-datatypes.md
+++ b/specification/gedcom-02-datatypes.md
@@ -37,7 +37,7 @@ They are represented as a string matching the same production as a tag,
 with the additional permission that standard enumerations may be integers.
 
 ```abnf
-stdEnum = stdTag / digit *tagchar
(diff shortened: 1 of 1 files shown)


dthaler:
As noted in issue #229: I would prefer option 1 more than allowing things like "2ND".   There's no demand for the permissiveness of "2ND" and it's not as good, in my view, as "2" or "SECOND", either of which are fine in option 1.   Since we can add permissiveness later, but can never remove permissiveness without a major version change, I'd argue for a less permissive fix now.

specification/gedcom-02-datatypes.md
```diff
@@ -33,14 +33,16 @@ The URI for the `Integer` data type is `xsd:nonNegativeInteger`.
 ## Enumeration
 
 An enumeration is a selection from a set of options.
-They are represented as a string matching the same production as a tag.
+They are represented as a string matching the same production as a tag,
+with the additional permission that standard enumerations may begin with a digit.
```
tychonievich, line 37:
```suggestion
with the additional permission that standard enumerations may be integers.
```

specification/gedcom-02-datatypes.md
```diff
@@ -33,14 +33,16 @@ The URI for the `Integer` data type is `xsd:nonNegativeInteger`.
...
 
 ```abnf
-Enum    = Tag
+stdEnum = stdTag / digit *tagchar
```
tychonievich, line 40:
```suggestion
stdEnum = stdTag / Integer
```

tychonievich: Less permissive option

Issue #234, opened 2022-10-18
Multiple translations of the same NOTE


dthaler:
Unlike every other use of TRAN in the spec, TRAN is {0:1} in the NOTE_STRUCTURE.
It should be {0:M} like other uses.

PR #235, opened 2022-10-18
Fix inconsistent cardinality of NOTE.TRAN

tychonievich:
See #234 for more

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1002,7 +1002,7 @@ means "no marriage had occurred as of March 24^th^, 1880"
 n NOTE <Text>                              {1:1}  g7:NOTE
   +1 MIME <MediaType>                      {0:1}  g7:MIME
   +1 LANG <Language>                       {0:1}  g7:LANG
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -1018,12 +1018,6 @@ Each `NOTE`.`TRAN` must have either a `MIME` or `LANG` substructure, and may hav
 
 See `SHARED_NOTE_RECORD` for advice on choosing between `NOTE` and `SNOTE`.
 
-:::note
-Although notes may be tagged with the language that they are written in, this specification does not provide a mechanism for distinguishing between notes with distinct content in distinct languages versus notes with the same content translated into distinct languages. It is expected that a future version of this specification will provide a mechanism for a single note to have multiple language translations.
```
dthaler, line 1022: Upon re-reading, I think the first sentence of this note is still valid.   That is, LANG is still {0:1} so if the NOTE <Text> contains a mix of two languages, there is no way for distinguishing which two languages are in the text.

tychonievich: I don't think that's what the note means. We could add a note like that, but this note is not that note.

Issue #236, opened 2022-10-23
TIME as a substructure of DATE


hartenthaler:
TIME is allowed in the standard as a substructure of DateValue. DateValue can be for example a DatePeriod:

> DateValue = date / DatePeriod / dateRange / dateApprox / ""

But what is the meaning of 
2 DATE FROM 1 JAN 2000 to 31 DEC 2000
3 TIME 12:34

btw: why is it "DatePeriod" with a capital "D" and not "datePeriod" like dateRange or dateApprox?

fisharebest:
> But what is the meaning of

Suppose I told you that my child was born at 12:34, in the year 2000...

tychonievich:
> btw: why is it "DatePeriod" with a capital "D" and not "datePeriod" like dateRange or dateApprox?

Per [their defining standard](https://www.rfc-editor.org/info/std68) ABNF rule names are case-insensitive. During the drafting of 7.0.0 we noticed that other standards seemed to use upper-case for externally visible rules and lower-case for internal-use-only rules, so we tried to follow that practice. Hence we used `DatePeriod` because it's used to define a datatype but `dateApprox` because it is only used to define part of `DateValue`.

That initial casing rule goal was difficult to formally define and follow. For example, `datePeriod` is used in the first paragraph of section 3.3.1; is that "externally visible" enough to deserve upper-case naming? We decided not, but also decided that this kind of corner case would make formal specification of the casing rules too difficult to define, which is why they are not called our explicitly in the specification.

Norwegian-Sardines:
Actually FROM/TO or datePeriod means that an event happened continually FROM a TO b, so a TIME would be stating that at TIME c every day FROM a TO b the event occurred.  I would say that while logical, this type of event would rarely occur, maybe you could say: “The captains wife walked down to the dock to check if his ship arrived very day at noon from the day he left until she died”

The issue is that some illogical date/time combinations can be produced but sometimes a logical statement such as in “A child was born in 2012 at 14:32” needs to be recorded.

tychonievich:
I agree with @hartenthaler that the specification does not define what it means to have a time under a variety of dates.

- Does `FROM <date> TO <date>` with `TIME <time>` mean the multi-day event both started and stopped at that time? Or was it a periodic event that took place each day at that time during that period?
[...]

PR #230, opened 2022-10-13
Add TIME.PHRASE

tychonievich:
Also a recommendation about `TIME`.`PHRASE` not being used with `g7:DATE-exact`

See #168 for the longer discussion about this topic.

PR #241, opened 2022-11-08
Pull DATE_VALUE into own section and add comments about time

tychonievich:
This is an effort to address #236 which observed that many types of `DateValue` payloads leave `DATE`.`TIME` under-defined.

It also adds a new section for the `<<DATE_VALUE>>` structure to create a good place to address this issue. If we like this, it will mean we'll need to refactor #230 to use this new section structure

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -540,7 +540,8 @@ n DATE <DateValue>                         {1:1}  g7:DATE
 ```
 
 A date, optionally with a time and/or a phrase.
-The `TIME` should only be provided if the `DATE` identifies a single day.
+If there is a `TIME`, it asserts that the event happened at a specific time on a single day.
+`TIME` should not be used with `DatePeriod` but may be used with other date types.
 
 :::note
 Previous versions of the spec failed to restrict `TIME` to single-day `DATE`s, but also did not define the meaning of such a time.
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -531,12 +531,33 @@ n CREA                                     {1:1}  g7:CREA
...
 
 ```gedstruct
-n DATE <DateValue>                         {0:1}  g7:DATE
+n DATE <DateValue>                         {1:1}  g7:DATE
   +1 TIME <Time>                           {0:1}  g7:TIME
   +1 PHRASE <Text>                         {0:1}  g7:PHRASE
+```
+
+A date, optionally with a time and/or a phrase.
+The `TIME` should only be provided if the `DATE` identifies a single day.
```
tychonievich, line 543:
```suggestion
If there is a `TIME`, it asserts that the event happened at a specific time on a single day.
`TIME` should not be used with `DatePeriod` but may be used with other date types.
```

specification/gedcom-03-datamodel.md
```diff
@@ -531,12 +531,34 @@ n CREA                                     {1:1}  g7:CREA
...
+If there is a `TIME`, it asserts that the event happened at a specific time on a single day.
+`TIME` should not be used with `DatePeriod` but may be used with other date types.
+
+:::note
+Previous versions of the spec failed to restrict `TIME` to single-day `DATE`s, but also did not define the meaning of such a time.
+Because of this, if an approximate date contains a time there is no way to know if the intent was that the day was unknown but the time was known or if the intent was that the time is likewise uncertain.
+
+The meaning of a time for a `DatePeriod` is likewise not defined.
+:::
+
```
tychonievich, line 552:
```suggestion
```

PR #251, opened 2022-11-22
Fix media type definition

### LINKED ISSUE #249: Possible mistake in definition of mt-char
In section 2.10 Media Type, `mt-char` is defined to include the space character `%x20`. But the corresponding production in [RFC 2045, section 5.1](https://www.rfc-editor.org/rfc/rfc2045#section-5.1) says:

>  token := 1*<any (US-ASCII) CHAR except SPACE, CTLs, or tspecials>

It seems unlikely that a space can be part of the name of a media type, subtype or attribute, as GEDCOM 7.0 appears to say.

### LINKED ISSUE #252: Are extension media types usable with g7:FORM?
The GEDCOM 7 spec says:
```
FORM (Format) g7:FORM
The media type of the file referenced by the superstructure. This should be a valid media type as
defined by [BCP 13](https://www.rfc-editor.org/info/bcp13). A [registry of media types](https://www.iana.org/assignments/media-types/media-types.xhtml)
is maintained publicly by the IANA.
```

However RFC 2045 defines media types that can be registered, and also defines extension types/subtypes that start with "x-":
```
content := "Content-Type" ":" type "/" subtype
                *(";" parameter)
                ; Matching of media type and subtype
                ; is ALWAYS case-insensitive.

     type := discrete-type / composite-type

     discrete-type := "text" / "image" / "audio" / "video" /
                      "application" / extension-token

     composite-type := "message" / "multipart" / extension-token

     extension-token := ietf-token / x-token

     ietf-token := <An extension token defined by a
                    standards-track RFC and registered
                    with IANA.>

     x-token := <The two characters "X-" or "x-" followed, with
                 no intervening white space, by any token>

     subtype := extension-token / iana-token

     iana-token := <A publicly-defined extension token. Tokens
                    of this form must be registered with IANA
                    as specified in [RFC 2048](https://www.rfc-editor.org/rfc/rfc2048).>
```

BCP 13 is only about _registered_ types, not extension types, per the title of the documents in BCP13.  That's why it always says "All registered media types MUST" etc, rather than "All media types MUST".

So currently section 3 of the GEDCOM spec implies, by saying "This should be a valid media type as defined by BCP 13", that only registered types are legal and "x-" types are illegal.  However, lower case "should" is ambiguous, and it's unclear whether that is the intent or not, especially since `<MediaType>` is defined in section 2 in a way that is not explicitly limited to registered types:
```
The media type data type represents the encoding of information in bytes
or characters, as defined in [RFC 2045](https://www.rfc-editor.org/info/rfc2045)
and [registered by the IANA](http://www.iana.org/assignments/media-types/).

The official grammar for media type is given in RFC 2045, section 5.1.
```

Also notably the legal syntax for registered types is more constrained (by RFC 6838) than the syntax for extension types.
As an example `~` (tilde) was apparently allowed by the RFC 2045 ABNF but disallowed for registered values.  So I think `x-a~b` is legal for an unregistered name but `a~b` is not legal in a registered name.

dthaler:
Fixes #249 
Fixes #252 

### DIFF
--- a/specification/gedcom-02-datatypes.md
+++ b/specification/gedcom-02-datatypes.md
@@ -295,26 +295,21 @@ The URI for the `Language` data type is `xsd:Language`.
 
 The media type data type represents the encoding of information in bytes or characters, as defined in [RFC 2045](https://www.rfc-editor.org/info/rfc2045) and [registered by the IANA](http://www.iana.org/assignments/media-types/).
 
-The official grammar for media type is given in RFC 2045, section 5.1.
-However, that document does not give stand-alone ABNF, instead referring to registration rules and describing some components in English.
-The programmatic parts of the media type grammar can be summarized as follows:
+The official grammar for media type is given in RFC 2045, section 5.1, and the ABNF was updated
+in [BCP 13](https://www.rfc-editor.org/info/bcp13).
 
 ```abnf
-MediaType    = mt-type "/" mt-subtype *(";" mt-parameter)
-
-mt-type      = mt-token
-mt-subtype   = mt-token
-mt-parameter = mt-attribute "=" mt-value
-mt-token     = 1*mt-char
-mt-attribute = mt-token
-mt-value     = mt-token / mt-qstring
-mt-char      = %x20-21 / %x23-27 / %x2A-2B / %x2D-2E ; not "(),/
-             / %x30-39 / %x41-5A / %x5E-7E           ; not :;<=>?@[\]
-
-mt-qstring   = %x22 *(mt-qtext / mt-qpair) %x22
-mt-qtext     = %x09-0A / %x20-21 / %x23-5B / %x5D-7E ; not CR "\
-mt-qpair     = "\" %x09-7E
+MediaType = type "/" subtype parameters
 ```
+where:
+* `type` and `subtype` are defined in [RFC 2045](https://www.rfc-editor.org/info/rfc2045)
+  section 5.1, and registered values (i.e., those not beginning with "x-") are further
+  constrained by the definitions in
+  [RFC 6838](https://www.rfc-editor.org/info/rfc6838), section 4.2, and
+* `parameters` is defined in [RFC 9110](https://www.rfc-editor.org/info/rfc9110),
+  section 5.6.6.  Note that the `parameters` definition here matches that used by HTTP
+  headers which permit whitespace around the ";" delimiter, whereas email headers in
+  RFC 2045 do not.
 
 The URI for the `MediaType` data type is `dcat:mediaType`.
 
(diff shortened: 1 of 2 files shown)


tychonievich (commented):
I believe this change is technically correct. However, it will mean that our ABNF will not longer be self-contained, potentially breaking tools that use the extracted ABNF to create a parser. We know that such tools exist (see e.g. #206). Is that a concern we should worry about?

dthaler:
> I believe this change is technically correct. However, it will mean that our ABNF will not longer be self-contained, potentially breaking tools that use the extracted ABNF to create a parser. We know that such tools exist (see e.g. #206). Is that a concern we should worry about?

As noted in the original github issue, the ABNF in the document was wrong.   This change takes the same approach as for language tags by referencing the correct ABNF elsewhere.

tychonievich:
> As noted in the original github issue, the ABNF in the document was wrong. This change takes the same approach as for language tags by referencing the correct ABNF elsewhere.

I understand and agree with that. My concern is about the *extracted* ABNF in [extracted-files/grammar.abnf](/FamilySearch/GEDCOM/blob/main/extracted-files/grammar.abnf), which is generated by concatenating all `abnf`-type code blocks in the markdown.

Currently, all rule names that appear in that extracted file are defined in that file. Language tags do this by not having an `abnf`-type code block at all: it refers to a single rule name form another standard in the test instead.

This pull request changes it so that there are three rule names (`type`, `subtype`, and `parameters`) that are used in the extracted file without being defined in it.

Is that change from self-contained to having-external-dependencies of the extracted file itself worth worrying about? If it is a worry, should we fix it by (a) refactoring the rule to not have an `abnf`-type code block, (b) adding special handling to remove it from the extracted file, or (c) copying the relevant rules from the referenced standards?

dthaler:
> Is that change from self-contained to having-external-dependencies of the extracted file itself worth worrying about? If it is a worry, should we fix it by (a) refactoring the rule to not have an `abnf`-type code block, (b) adding special handling to remove it from the extracted file, or (c) copying the relevant rules from the referenced standards?

I would only copy the relevant rules from the referenced standards if we do the same for language tags.

tychonievich:
Discussed 2022-11-29
We are leaning towards leaving dangling references in the extracted ABNF. @DavidMStraub, will that be a problem for your Python parser? If so, our next preference is to extract the ABNF rules from the relevant RFCs into the extracted file.

DavidMStraub:
>  @DavidMStraub, will that be a problem for your Python parser? 

Thanks for asking. While it would indeed prevent the automated conversion of the ABNF rules to regular expressions that I needed to create the parser, It's not a practical problem right now as I already have them. I can manually fix the regex for media type.

Just a curiosity, would it be an option to define `type = Special` etc., since it's also a special type (not further specified here, but elsewhere)? This would avoid the dangling reference.

PR #270, opened 2023-01-31
Note that `AFT` dates may not indcate an event ocurred

### LINKED ISSUE #264: Disconnect between spec and common use of event occurance
The spec currently says that any event with a DATE substructure is an assertion that the event did occur. But as noted in the comments on #258 , that is not the intuitive meaning for some date periods like

```gedcom
1 DEAT
2 DATE AFT 1800
```

We could

1. leave it as is
2. make a breaking change to align the definition with the intuitive meaning (i.e. in version 8.0)
3. make a breaking change to remove the "DATE means it occurred" concept entirely and always require `Y` for that
4. add a clarifying note explaining that intuition and the spec have not agreed for some decades and thus concluding that it did or did not occur in these cases is unwise

tychonievich:
Resolves #264

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -1221,6 +1221,30 @@ An event structure asserts the event did occur if any of the following are true:
     ````
     
     </div>
+    
+    <div class="note">
+    
+    The presence of a `DATE` indicating an event occurred was introduced to in version 5.4 (1995), and is generally followed by applications.
+    However, it is common for users to enter a date range with no end without intending to indicate that the event occurred.
+    For example, pre 7.0 files sometimes used
+    
+    ````gedcom
+    1 NATU
+    2 DATE AFT 1800
+    ````
+    
+    to mean what 7.0 encodes as
+    
+    ````gedcom
+    1 NOT NATU
+    2 DATE BEF 1800
+    ````
+    
+    without intending to imply that `NATU` ever did actually occur.
+    Because this is a "sometimes used" rather than a "formally means" situation,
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -1221,6 +1221,30 @@ An event structure asserts the event did occur if any of the following are true:
     ````
     
     </div>
+    
+    <div class="note">
+    
+    That a `DATE` indicates an event occurred was introduced to in version 5.4 (1995), and is generally followed by applications.
```
dthaler, line 1227: I can't parse this sentence.  Not sure what "to in" means, and also reads awkward starting the sentence with "That".

specification/gedcom-03-datamodel.md
```diff
@@ -1221,6 +1221,30 @@ An event structure asserts the event did occur if any of the following are true:
...
+    to mean what 7.0 encodes as
+    
+    ````gedcom
+    1 NOT NATU
+    2 DATE TO 1800
+    ````
+    
+    without intending to imply that `NATU` ever did actually occur.
+    Because this is a "sometimes used" rather than a "formally means" situation,
+    it is likely that data using 5.x "after meaning not before" *de facto* pattern will be transferred as-is into 7.0 and persist in files for the foreseeable future.
```
dthaler, line 1245: Does this suggest that the [migration guide](https://gedcom.io/migrate/) should suggest converting the former (NATU.DATE AFT 1800) to the latter ("NOT NATU.DATE TO 1800")?

specification/gedcom-03-datamodel.md
tychonievich, line 1227:
```suggestion
    The presence of a `DATE` has indicated an event occurred since version 5.4 (1995), and is generally followed by applications.
```

PR #274, opened 2023-02-09
Resolve #256 by adding PEDI.DATE

### LINKED ISSUE #256: FAMC.PEDI.DATE
There are situations where individuals are foster children in different families at different times. In order to also document this in GEDCOM, a time stamp is required. Currently, this is not provided by the FAMC.PEDI structure.
+1 FAMC @<XREF:FAM>@
+2 PEDI <Enum>
+3 PHRASE <Text>

To this current structure a
+3 DATE <DatePeriod>
should be added.

Currently the PHRASE could be used by a "FROM TO" Text.
Or ist a "+3 _DATE FROM ... TO ..." a better solution?

tychonievich:
Also makes `PEDI` `{0:M}` not `{0:1}`, which could have use (e.g. when a child was both adopted by and fostered to the same family) even without knowing dates.

resolves #256


lynchrs3 (approved):
I support this change.

dthaler (approved):

PR #275, opened 2023-02-09
Resolve #273 by removing BCE from dateRestrict

### LINKED ISSUE #273: Date, dateRestrict, epoch: confusing constraints
dateRestrict contains tags, which must not be recognized as month, calendar or epoch (mentioned in additional contraints). dateRestrict however contains "BCE" which is a valid tag in epoch. This seems confusing.

tychonievich:
Also adds a recommendation that months and epochs use distinct tags


lynchrs3 (approved):
I support the removal of BCE

dthaler (approved):

PR #283, opened 2023-02-23
Add SUBM.ALIA

### LINKED ISSUE #268: Link SUBM and INDI
It is common for submitters to include an INDI representing themselves, and common for applications to use that information to inform the user interface. Support for storing that relationship should be included in the spec.

The proposal in a comment in #191 was to have a pointer substructure of SUBM pointing to the relevant INDI. Presumably this would have cardinality `{0:1}` to support the common user-interface deisngs that use this information. Two tags have been proposed for this: `SUBM`.`INDI` and `SUBM`.`ALIA`; we could also create a new tag for this case.

_Extracted from a post by @fisharebest in https://github.com/FamilySearch/GEDCOM/discussions/191#discussioncomment-4743752_

tychonievich:
Intended to resolve #268

ALIA was chosen because it was used in an example in the 5.5.1 spec, and the meaning is similar to INDI.ALIA. {0:1} was chosen to support the "root individual" and related ideas.

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -447,6 +447,7 @@ n @XREF:SUBM@ SUBM                         {1:1}  g7:record-SUBM
   +1 WWW <Special>                         {0:M}  g7:WWW
   +1 <<MULTIMEDIA_LINK>>                   {0:M}
   +1 LANG <Language>                       {0:M}  g7:SUBM-LANG
+  +1 ALIA @<XREF:INDI>@                    {0:1}  g7:SUBM-ALIA
   +1 <<IDENTIFIER_STRUCTURE>>              {0:M}
   +1 <<NOTE_STRUCTURE>>                    {0:M}
   +1 <<CHANGE_DATE>>                       {0:1}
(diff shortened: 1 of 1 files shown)


dthaler (approved):

tychonievich:
Would also resolve #191

lynchrs3 (approved):
I accept this change.

chronoplexsoftware:
This is an interesting problem.  This change would allow identification of the root person of one graph, without relying on ordering.  However within a single GEDCOM file, it is possible to have multiple graphs, and it is desirable to identify a root person for each graph such that users can jump to the root person of each graph easily.

To support multiple root people, we currently ensure that the root person in a given graph appears in the GEDCOM file before anyone else in that graph.  This avoids the need for the use of extensions, but comes with the risk that the root person will be lost if other applications regenerate the file and change the ordering.  A solution which also considers the above would be desirable.

tychonievich:

Because many files (e.g. one-place studies) have many disjoint trees, this is not a solution to the root person of tree problem (#191). However, it does solve the connection between users (SUBM) and historical persons (INDI) and hence does resolve #268. A separate solution will be needed for the root person problem.

specification/gedcom-03-datamodel.md
```diff
@@ -1421,6 +1422,19 @@ Others distribute events and attributes between `INDI` records mutually linked b
...
+Indicates that the user described by the superstructure
+is the same person as the individual described by the referenced `INDI`.
+
+A `SUBM` may only have a single `INDI` linked in this way
+to simplify implementation of algorithms that depend on the `INDI` representing a particular user.
+Applications using `g7:ALIA` to represent a person with multiple `INDI` records
+should select or generate a representative `INDI` for the `SUBM`.`ALIA`.
+
+Applications requiring a "root individual" for display or navigational purposes
+should use the `HEAD`.`SUBM`'s `ALIA` if that is available.
```
tychonievich, line 1436:
```suggestion
```

PR #295, opened 2023-03-09
Add URIs for all pseudostructures

tychonievich:
- `HEAD` as `g7:HEAD`, replacing "HEAD pseudostructure" in extracted files
- `TRLR` as `g7:TRLR`
- `CONT` as `g7:CONT`; also add that it is removed during parsing and never appears in parses datasets
- `CONC` as `g7:CONC`; also document that this had meaning in pre-7 versions and will not appear in 7.0 nor be re-used for a different stucture in the future

### DIFF
--- a/build/uri-def.py
+++ b/build/uri-def.py
@@ -162,7 +162,7 @@ def new_key(val, d, *keys, msg=''):
 
 def parse_gedstruct(txt, rules, dtypes):
     """Reads through all gedstruct blocks to find payloads, substructures, and superstructures"""
-    sup,sub,payload = {}, {}, {}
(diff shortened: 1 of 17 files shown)


tychonievich:
I toyed with different URIs; thus far g7:TAG format has been reserved for substructures. The alternatives I almost put in this PR are

`g7:Header`
`g7:Trailer`
`g7:pseudo-CONT`
`g7:pseudo-CONC`

tychonievich:
Discussed in steering committee; we should not give a URI for CONC in v7 because it doesn't appear in version 7; the text for CONC should be moved up to the section on lines isntead of the list of structures.

PR #302, opened 2023-04-13
Add NO NAME

### LINKED ISSUE #246: NO NAME
The `NO` tag can only be used with events.  e.g. `NO MARR`.

If an individual was not named, then we have no way to indicate that no name existed.
For example, I have death records that just state "boy" or "girl", rather than a name.

Should the `NO` tag also allow `NAME`?

tychonievich:
This involves several parts. Because `NO NAME` and `NO IMMI` would appear in the same context, they must have the same structure type. That means the name "`NON_EVENT_STRUCTURE`" is not ideal anymore (this changes it to `NEGATIVE_ASSERTION`) and that the enumeration set should be renamed from `enumset-EVEN` to `enumset-NO`.

This also patches a gap in the previous spec whereby `NO BIRT` could be a substructure of `FAM`, but does so with a "should" not a "must" to avoid breaking backwards compatibility.

Resolves #246

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -2906,12 +2906,16 @@ See [Events] and [Attributes].
 
 ### `g7:enumset-NO` {.unlisted .unnumbered}
 
-This set contains various structure type's tags:
-every event-type tag name except the generic `EVEN` tag,
-as well as `NAME`, `GIVN`, and `SURN`,
-each referring to the information described by the corresponding structure type.
-See [Events], `g7:GIVN`, `g7:INDI-NAME`, and `g7:SURN`.
+This set contains various structure types, with the same tags and URIs as the respective structures:
 
+- All individual and family event types except `EVEN`
+
+- The personal name type `g7:INDI-NAME`
+
+- The name parts `g7:GIVN` and `g7:SURN`
+
+Values should be appropriate to the context of the containing structure,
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -2902,6 +2904,15 @@ See [Events] and [Attributes].
 | `VIDEO`      | Motion picture recording          |
 | `OTHER` | A value not listed here; should have a `PHRASE` substructure |
 
+### `g7:enumset-NO` {.unlisted .unnumbered}
+
+This set contains various structure type's tags:
```
dthaler, line 2909: Can't parse grammar of line 2909.  Reword?

dthaler (commented):
What about 
```
NO SSN
NO NATI
```
?

The latter I think would require allowing `TYPE` as a substructure of `NO` which would be good in my view anyway.  E.g., to denote that someone was not adopted.

specification/gedcom-03-datamodel.md
```diff
@@ -2161,8 +2162,14 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 
 #### `NO` (Did not happen) `g7:NO`
```
tychonievich, line 2163:
```suggestion
#### `NO` (Negative assertion) `g7:NO`
```

PR #307, opened 2023-05-05
Short, labeled EXID-TYPE

### LINKED ISSUE #23: Add display-name for URIs
EXID.TYPE has a URI payload, which is good for its identifying purpose but less good to show users. For example,

```gedcom
2 PLAC Querfurt
3 EXID QUEURTJO51TJ
4 TYPE https://gov.genealogy.net/item/show/
```

should be displayed to the user as "`QUEURTJO51TJ`" being the "GOV id" or the like; simply using the URL host name  is not ideal.

Option 1: add `EXID`.`TYPE`.`TITL`. Simple change, local to where the data is displayed, but means the data has to be repeated each EXID.

Option 2: add something like a HEAD.DISPLAYNAME structure where the display name for each URI can be specified just once (possibly localized with LANG tags). More complicated change, but potentially more versatile; for example, it could also be used to give display names for enumeration values.

Option 3: extend the SCHMA to allow specifying URIs of things other than tags and providing more details about them, such as display names. If well designed this makes for the most orderly solution, but there are many options to consider in designing this change.

### LINKED ISSUE #177: EXIDs cause up to a 5x size increase
Let's take the example in the migration guide:

> 1 RIN 9876

This is 11 bytes (including a newline).   The migration guide says to convert to

> 1 EXID 9876
> 2 TYPE https://gedcom.io/terms/v7/RIN#MySystem

which is 57 bytes, over 5x larger than using RIN.   The same size increase applies to, say, FamilySearch Source Description IDs compared to using an extension tag for it.

GEDCOM file size can affect many things, including disk usage, import/export time, bandwidth to transfer files, and there may be size maximums that get hit such as maximum size of attachments supported by some email systems.   As such, FamilySearch GEDCOM 7 files may be much less performant or usable than GEDCOM 5.5.1 when use of external IDs are ubiquitous in a file.

The GEDZIP format can alleviate some of this, at the expense of even longer import/export time, and the loss of human readability of attachments.

Should we do anything about this in a future version?

In the meantime, applications that store make heavy use of external IDs may either use extension tags instead of EXIDs, or even just stick with GEDCOM 5.5.1 if they just use RIN/RFN/AFN.

tychonievich:
Resolves #177 by adding extTag option for EXID.TYPE payload

Resolves #23 by adding suggeted labels for all documented extTags, including EXID.TYPEs but also extension structures. Because the labels in one dataset could differ from those suggested in official documentation, also clarifies that in-dataset labels are user-preference suggestions.

### DIFF
--- a/specification/gedcom-02-datatypes.md
+++ b/specification/gedcom-02-datatypes.md
@@ -317,6 +317,27 @@ where:
 
 The URI for the `MediaType` data type is `dcat:mediaType`.
 
+## Shortened URI
+
+A URI, possibly shorted and documented using a [documented extension tag](#extension-tags).
+
+```abnf
+ShortURI = extTag / URI
+```
+
+where:
+
+* `URI` is defined in [STD 66](https://www.rfc-editor.org/info/std66) section 3
+  as what is commonly called an "absolute URI"; notably, it always includes a colon.
+
+* Any `extTag` used shall be documented in the [schema] as a [documented extension tag](#extension-tags).
+  As with other documented extension tags, the tag may be changed without changing the meaning of the data
+  provided that the URI it represents does not change.
+
+Using `extTag` is preferred when the same URI would occur repeatedly in the dataset
+or when substructures of `g7:TAG` are used in the documented extension tag definition to provide additional information.
+
+
 ## Special
 
 The special data type is a string conforming to a case-specific standard or constraints. The constraints on each special data type instance are either unique to that structure type or are not simply expressed.
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -2003,7 +2003,7 @@ FamilySearch GEDCOM 7.0 reintroduced it with the name `INIL` for consistency wit
 #### `LABEL` (Label) `g7:LABEL`
 
 A recommended short label to use in displaying information described by the superstructure to the user.
-Multiple labels may be provided for the same superstructure;
+Multiple labels may be provided for the same superstructure.
 As with other structures, those with the same qualifying substructures (such as `g7:LANG`) are in user preference order.
 
 #### `LANG` (Language) `g7:LANG`
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -134,6 +134,8 @@ n HEAD                                     {1:1}  g7:HEAD
      +2 VERS <Special>                     {1:1}  g7:GEDC-VERS
   +1 SCHMA                                 {0:1}  g7:SCHMA
      +2 TAG <Special>                      {0:M}  g7:TAG
+        +3 LABEL <Text>                    {0:M}  g7:LABEL
+           +4 LANG <Language>              {0:1}  g7:LANG
   +1 SOUR <Special>                        {0:1}  g7:HEAD-SOUR
      +2 VERS <Special>                     {0:1}  g7:VERS
      +2 NAME <Text>                        {0:1}  g7:NAME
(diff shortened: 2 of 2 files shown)


specification/gedcom-02-datatypes.md
```diff
@@ -317,6 +317,27 @@ where:
 
 The URI for the `MediaType` data type is `dcat:mediaType`.
 
+## Shortened URI
+
+A URI, possibly shorted and documented using a [documented extension tag](#extension-tags).
```
dthaler, line 322:
```suggestion
A URI, possibly shortened and documented using a [documented extension tag](#extension-tags).
```

specification/gedcom-02-datatypes.md
```diff
@@ -317,6 +317,27 @@ where:
...
+```abnf
+ShortURI = extTag / URI
+```
+
+where:
+
+* `URI` is defined in [STD 66](https://www.rfc-editor.org/info/std66) section 3
+  as what is commonly called an "absolute URI"; notably, it always includes a colon.
+
+* Any `extTag` used shall be documented in the [schema] as a [documented extension tag](#extension-tags).
```
dthaler, line 333: make `[schema]` be a link?

specification/gedcom-03-datamodel.md
```diff
@@ -1998,6 +2000,12 @@ A [Latter-Day Saint Ordinance](#latter-day-saint-ordinances).
 See also `LDS_INDIVIDUAL_ORDINANCE`.  Previously, GEDCOM versions 3.0 through 5.3 called this `WAC`; it was not part of 5.4 through 5.5.1.
 FamilySearch GEDCOM 7.0 reintroduced it with the name `INIL` for consistency with `BAPL`, `CONL`, and `ENDL`.
 
+#### `LABEL` (Label) `g7:LABEL`
+
+A recommended short label to use in displaying information described by the superstructure to the user.
+Multiple labels may be provided for the same superstructure;
```
dthaler, line 2006:
```suggestion
Multiple labels may be provided for the same superstructure.
```
Use period because 2006 starts capitalized.

specification/gedcom-02-datatypes.md
albertemmerich: I am not shure whether this solution leads to a way without extension tags in next minor/major. As the implementation in applications is not easy (so far they can build the complete link to the EXID directly), I would prefer a solution which remains very similar in following versions of the standard. Is there any idea how a solution without extension tags could work?
The proposed solution would work, but as the size of files is not very limiting, I see too much work for implementation compared to the temporary result.

specification/gedcom-02-datatypes.md
tychonievich: With the pandoc, the tool used to render this markdown, `[schema]` will be automatically interpreted as `[schema](#schema)`. I'm fine with replacing it with gfm-compatible markdown or leaving as is

dthaler:
@tychonievich PR needs to be rebased to resolve conflicts

specification/gedcom-02-datatypes.md
dthaler: github doesn't render `[schema]` as a link so would recommend replacing it.

specification/gedcom-03-datamodel.md
```diff
@@ -134,6 +134,8 @@ n HEAD                                     {1:1}  g7:HEAD
      +2 VERS <Special>                     {1:1}  g7:GEDC-VERS
   +1 SCHMA                                 {0:1}  g7:SCHMA
      +2 TAG <Special>                      {0:M}  g7:TAG
+        +3 LABEL <Text>                    {0:M}  g7:LABEL
+           +4 LANG <Language>              {0:1}  g7:LANG
```
dthaler, line 138: Shouldn't this label be in the YAML file (where it already exists)?

specification/gedcom-03-datamodel.md
```diff
@@ -1998,6 +2000,12 @@ A [Latter-Day Saint Ordinance](#latter-day-saint-ordinances).
 See also `LDS_INDIVIDUAL_ORDINANCE`.  Previously, GEDCOM versions 3.0 through 5.3 called this `WAC`; it was not part of 5.4 through 5.5.1.
 FamilySearch GEDCOM 7.0 reintroduced it with the name `INIL` for consistency with `BAPL`, `CONL`, and `ENDL`.
 
+#### `LABEL` (Label) `g7:LABEL`
+
+A recommended short label to use in displaying information described by the superstructure to the user.
+Multiple labels may be provided for the same superstructure.
+As with other structures, those with the same qualifying substructures (such as `g7:LANG`) are in user preference order.
```
dthaler, line 2007: What does it mean to have a non-preferred label?

tychonievich:
As this PR did not garner support in the steering committee and some of its provisions are outdated by the addition of the GEDCOM-registries repo, I am withdrawing this proposal.

Issue #316, opened 2023-05-27
MEDI ELECTRONIC ambiguities


dthaler:
`ELECTRONIC` is ambiguous since could apply to an online source or a digital source in personal possession (e.g., floppy disk, DVD, CD-ROM, hard drive, flash drive, etc.)

If you have an audio recording on a CD, is it `AUDIO` or `ELECTRONIC`?
If you have a digital photograph, is it `PHOTO` or `ELECTRONIC`?

Also, CD-ROM is another ambiguity.  The fact that a source is from a CD-ROM is important in Evidence Explained since it explicitly appears in the citation text.  See EE pages 283, 378, 392, 509, 711, and 799 as examples.

https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#g7enumset-medi has values for, among other things: `ELECTRONIC` and `OTHER`.  It is unclear what to use for a CD-ROM, such as a book on CD-ROM (`BOOK` is defined as only a "bound" book).

`OTHER` with a `PHRASE` of CD-ROM is of course usable but may not be reliably parsed by programs due to the various forms it might occur in ("CD", "CDROM", "CD-ROM", etc.)

Option 1 (possible for 7.1): add an enumerated value `MEDI CDROM` for it, just like ones exist for microfilm, microfiche, etc.
Option 2: add a Note (in 7.0) with an example using `PHRASE CD-ROM` with `MEDI ELECTRONIC`.
Option 3: add a Note with an example using `PHRASE CD-ROM` with `MEDI OTHER`.

Norwegian-Sardines:
A digital artifact (the definition used for "Electronic") actually refers to a any item in a digital form, magazine, newspaper, music, photo, Facebook.  In Library Science we often refer to things as "Born-Digital".  

Yale University says: 
>Born-digital items are materials that are created in a digital format. This includes websites, email, digital photographs, electronic records, and more. Born-digital items are distinct from analog items that are subsequently digitized, such as paper manuscripts or photographs.

So any item on the list could be digitized (be "electronic") even a "bound book".

[...]

PR #317, opened 2023-06-22
Recommend CALN.MEDI identify first medium

tychonievich:
A partial resolution of #316, based on the conversation about CALN.MEDI which can be represented as a recommendation.
Does not address the proposal to deprecate ELECTRONIC and add more specific terms in its place as that would require a new minor version.

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -2100,6 +2100,13 @@ See also `FAMILY_EVENT_STRUCTURE`.
 
 An enumerated value from set `g7:enumset-MEDI` providing information about the media or the medium in which information is stored.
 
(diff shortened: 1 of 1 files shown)


dthaler (approved):

specification/gedcom-03-datamodel.md
```diff
@@ -2100,6 +2100,13 @@ See also `FAMILY_EVENT_STRUCTURE`.
 
 An enumerated value from set `g7:enumset-MEDI` providing information about the media or the medium in which information is stored.
 
+When `MEDI` is a substructure of a `g7:CALN`, it is recommended that its payload describes the medium directly found at that call number rather than a medium from which it was derived.
```
dthaler, line 2103: This is fine, but can we also add a note saying that `MEDI` as a substructure of a `g7:FORM` is not recommended to use and may go away in a future version of GEDCOM?

Issue #314, opened 2023-05-25
Rename `NATI` to match its definition


tychonievich:
The structure with the `NATI` tag is defined as

| Version | Name | Definition |
|-|-|-|
| 5.0, 5.3 | Nationality | The national heritage of an individual. |
| 5.4, 5.5, 5.5.1 | National or Tribal Origin | The person's division of national origin or other folk, house, kindred, lineage, or tribal interest. Examples: Irish, Swede, Egyptian Coptic, Sioux Dakota Rosebud, Apache Chiricawa, Navajo Bitter Water, Eastern Cherokee Taliwa Wolf, and so forth. |
| 7.0 | Nationality | An individual’s national heritage or origin, or other folk, house, kindred, lineage, or tribal interest. |

Despite the broadening definition over time, the name for the tag was reverted to 5.3's name with 7.0. I propose changing that name (only, not the definition) to one that batter captures the meaning of the structure, such as
 
- **people**, in the older sense when "people" and "persons" were distinct concepts
- **heritage**, though the English of this is much too broad, including things like inheritances
- **birth affiliation**, except you can be inducted into a tribe or naturalized into a nationality
- **affiliation**, except many affiliations (like clubs and political parties) are not encoded with NATI now
- **national affiliation / tribal identity**, chosen so its initials spell the tag NATI

None of these names strike me as ideal. Any preferences or suggestions?

Norwegian-Sardines:
I like to think of these as Cultural Heritages, Cultural Affiliations, Cultural Identity.

The examples are Cultural in nature, and looks to how an individual Identifies themself, which may not be in total just using national boarders or even nations that exist on the map today.  A person could be German and not live in Germany, or Roman and live in what became England.

dthaler:
@tychonievich Can this issue be closed or was there anything else not covered by PR #319?

dthaler:
Closing per discussion March 7, 2024

PR #319, opened 2023-06-22
NATI covers several gx fact types

tychonievich:
Update five terms that are subsumed by NATI:

- Clan
- Ethnicity
- Heimat
- Nationality
- Tribe

Note: I'm unsure of "Heimat" as it is not a term used in my native culture, but the references I could find online seemed to indicate it was subsumed by NATI.

See also the partially-related #314 which proposes changing the one-word title of the NATI structure type in the spec to better match its definition.

### DIFF
--- a/attribute-event-requests/INDI-attribute.md
+++ b/attribute-event-requests/INDI-attribute.md
@@ -10,11 +10,11 @@ Guides for using this document can be found in the associated [README.md](README
 | | | [Affiliation](#affiliation) | |
 | `CAST` | 5.0 | [Caste](#caste) | |
 | \*     | 5.0 | [Childless](#childless) | encoded as `NCH 0` |
-| | | [Clan](#clan) | |
+| | | [Clan](#clan) | perhaps `NATI`? |
 | \* | 3.0 | [Died Before Eight](#died-before-eight) | encoded as `DEAT`.`AGE <8y` |
 | `EDUC` | 4.0 | Education | |
-| | | [Ethnicity](#ethnicity) | |
-| | | [Heimat](#heimat) | |
+| | | [Ethnicity](#ethnicity) | perhaps `NATI`? |
+| | | [Heimat](#heimat) | perhaps `NATI`? |
 | `IDNO` | 5.3 | [Identifying Number](#identifying-number) | |
 | | | [Life Sketch](#life-sketch) | |
 | | | [Military Service](#military-service) | |
@@ -30,7 +30,7 @@ Guides for using this document can be found in the associated [README.md](README
 | \*     | 5.0 | [Single](#single) | encoded as `NMR 0` |
 | `SSN`  | 5.3 | Social Security Number | the USA's `IDNO` equivalent |
 | `TITL` | 3.0 | [Title](#title) | |
-| | | [Tribe](#tribe) | |
+| | | [Tribe](#tribe) | perhaps `NATI`? |
 
 # Details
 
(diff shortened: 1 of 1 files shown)


dthaler (approved):

dthaler:
Discussion 6/29/2023:
Heimat always refers to a place, not an adjective for a person from a place or lineage.
By analogy, an American's Heimat might be "America" or "Washington" or "Seattle", not "American" or "Washingtonian" or "Seattlite".  Would NATI be used with place names, or only the latter type of adjectives?  If the latter, then Heimat does not belong in this set.

attribute-event-requests/INDI-attribute.md
```diff
@@ -10,16 +10,16 @@ Guides for using this document can be found in the associated [README.md](README
...
 | `CAST` | 5.0 | [Caste](#caste) | |
 | \*     | 5.0 | [Childless](#childless) | encoded as `NCH 0` |
-| | | [Clan](#clan) | |
+| `NATI` | 5.4 | [Clan](#clan) | one of several types of `NATI` |
 | \* | 3.0 | [Died Before Eight](#died-before-eight) | encoded as `DEAT`.`AGE <8y` |
 | `EDUC` | 4.0 | Education | |
-| | | [Ethnicity](#ethnicity) | |
-| | | [Heimat](#heimat) | |
+| `NATI` | 5.4 | [Ethnicity](#ethnicity) | one of several types of `NATI` |
+| `NATI` | 5.4 | [Heimat](#heimat) | one of several types of `NATI` |
```
tychonievich, line 17:
```suggestion
| | | [Heimat](#heimat) | |
```

attribute-event-requests/INDI-attribute.md
```diff
@@ -180,11 +182,12 @@ Found in the following historical records:
 
 The most closely related structures are:
 
-- 
+- `NATI`: Defined as "an individual’s national heritage or origin, or other folk, house, kindred, lineage, or tribal interest." While "heimat" is not specifically included, it is related to "origin" and "house" which are.
```
tychonievich, line 185:
```suggestion
- `NATI`: Defined as "an individual’s national heritage or origin, or other folk, house, kindred, lineage, or tribal interest." By contrast "heimat" refers to a geographical region a person is from. That could overlap with a national or other regional identity, but need not do so in general.
```

attribute-event-requests/INDI-attribute.md
```diff
@@ -180,11 +182,12 @@ Found in the following historical records:
...
 
 Related proposals include
 
 - Ethnicity: ethnicities typically have some geographic association, but generally much larger and less fine-grained than a heimat
+- Nationality: generally broader than heimat in what region it covers
```
tychonievich, line 190:
```suggestion
- Nationality: nations are often larger regions than heimats, and may refer to nations that a person belongs to but is not geographically from
```

tychonievich (commented):
Changes to remove heimat from the NATI family

Norwegian-Sardines:
This can probably be closed, but I’m still not convinced that Heimat should not be part of this concept.  My understanding is that it is not equivalent to the place but that it is indeed the association to that place, one “homeland”!  The origin of the term as described in https://en.wikipedia.org/wiki/Heimat#Origin_of_the_term
I would use this to identify the region in German (and equally in Norway) my ancestors came from!  

This concept may have changed in more modern times as national identity in the world became more prevalent over regional identity the use of this term may have changed.

But that being said I would not use NATI to identify either Seattle or Seattlite. But I would use it to identify a person as Bavarian, or a Farm name, both still places!

Issue #315, opened 2023-05-26
How to best include/define `TYPE` enumerations for Attributes and Events


Norwegian-Sardines:
Part of creating a better set of Attributes and Events to cover aspects of both Genealogy and Family History into a future GEDCOM is to do a better job of bringing together "like" attributes and events under the same category.

This might include for example: combining multiple states of Marriage (Common Law, Religious, Civil), Burial (Cremation, Inhumation, Crypt, At Sea), Death (Natural, Murder, Suicide, Combat) or expanding attributes that may be defined as an "unstructured list" such as Physical Description `DSCR`.

One issue that we all have is that we can't know all states of a particular event/attribute or in the case of an unstructured list any specific information and internationalization.

The current structure of all Attributes (except `DSCR`) is the attribute tag (for example `EDUC`) to follow the tag with text "Masters of Science".

On the other hand Events currently are limited to the "exact meaning of the tag", as more Family History tags get added (various Military Related events, different type of personal relationships, burial type, death types) the need to use the `TYPE` tag to extend the meaning is warranted to reduce the number and better categorize the event.

A GEDCOM could look like this:
```` {.gedstruct .long}
1 MARR
2 TYPE Common Law
````

The question is:  How to define in the document the enumeration and the possibility of having an "other" and phrase for concepts not thought about today?

GEDCOM v7 currently uses enumeration sets for multiple tags, for example:

### `g7:enumset-MEDI`

| Value        | Meaning                           |
| :----------- | :-------------------------------- |
| `AUDIO`      | An audio recording                |
| `BOOK`       | A bound book                      |
| `CARD`       | A card or file entry              |
| `ELECTRONIC` | A digital artifact                |
| `FICHE`      | Microfiche                        |
| `FILM`       | Microfilm                         |
| `MAGAZINE`   | Printed periodical                |
| `MANUSCRIPT` | Written pages                     |
| `MAP`        | Cartographic map                  |
| `NEWSPAPER`  | Printed newspaper                 |
| `PHOTO`      | Photograph                        |
| `TOMBSTONE`  | Burial marker or related memorial |
| `VIDEO`      | Motion picture recording          |
| `OTHER` | A value not listed here; should have a `PHRASE` substructure |

With GEDCOM Structure definition:

```` {.gedstruct .long}
+1 FILE <Special>                            {1:M}  g7:FILE
    +2 FORM <MediaType>                      {1:1}  g7:FORM
[...]

PR #322, opened 2023-07-06
Add KIND enumerated values for MARR and BURI

tychonievich:
Conversation draft of what a KIND might look like

See #315 for more on this idea and why it's KIND instead of TYPE. Each `KIND` has its own structure type so that it can have its own enumeration set.

This draft uses `KIND <Enum> {0:1}`, but we could allow several KINDs, either by `KIND <List:Enum> {0:1}` or `KIND <Enum> {0:M}`. I think single-kind is simpler, but because of that didn't include the `GROUP` kind proposed in <https://github.com/FamilySearch/GEDCOM/issues/315#issuecomment-1564557254> as it would presumably usually be paired with another kind.

I found deciding what was and wasn't a kind of burial or marriage to be tricky. For example,

- I included sky burials, even though those are sometimes classes as a body preparation step like cremation or embalming instead, but didn't include embalming or mummifying.
- I didn't include cohabitating/partnership/companionship marriage kinds because I thought some researchers might class those as being closer to a `NO MARR` than a `MARR`, but I'm not sure that was the right decision.
- I found two different definitions of common-law marriage, so I split those into `COMMON_LAW` and `INFORMAL`, but I'm not confident that was the right split

Before discussing those topics, though, I hope this PR can facilitate discussion of the following questions:

1. If we add KIND in 7.1, should it be
    1. `{0:1}` or
    2. `{0:M}` or
    3. vary by event type (e.g. marriage is `{0:M}` but burial is `{0:1}`)?
2. Is the "use TYPE instead of PHRASE" the right decision? Is it worded clearly enough? Is it OK to use `OTHER` with `PHRASE` recommended most places but `TYPE` recommended for `KIND OTHER`?
3. If we like this idea, how should we decide which events get KINDs?

### DIFF
--- a/specification/gedcom-03-datamodel.md
+++ b/specification/gedcom-03-datamodel.md
@@ -680,6 +680,7 @@ n MARL [Y|<NULL>]                          {1:1}  g7:MARL
 |
 n MARR [Y|<NULL>]                          {1:1}  g7:MARR
   +1 TYPE <Text>                           {0:1}  g7:TYPE
+  +1 KIND <Enum>                           {0:1}  g7:MARR-KIND
   +1 <<FAMILY_EVENT_DETAIL>>               {0:1}
 |
 n MARS [Y|<NULL>]                          {1:1}  g7:MARS
@@ -851,6 +852,7 @@ n BLES [Y|<NULL>]                          {1:1}  g7:BLES
 |
 n BURI [Y|<NULL>]                          {1:1}  g7:BURI
   +1 TYPE <Text>                           {0:1}  g7:TYPE
+  +1 KIND <Enum>                           {0:1}  g7:BURI-KIND
   +1 <<INDIVIDUAL_EVENT_DETAIL>>           {0:1}
 |
 n CENS [Y|<NULL>]                          {1:1}  g7:INDI-CENS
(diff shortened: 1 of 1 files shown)


specification/gedcom-03-datamodel.md
```diff
@@ -1998,6 +2000,18 @@ A [Latter-Day Saint Ordinance](#latter-day-saint-ordinances).
...
 
+#### `KIND` (Kind) `g7:BURI-KIND`
+
+An enumerated value from set `g7:enumset-BURI-KIND` indicating what was done with the remains of the deceased.
+
+Note `KIND` does not have a `g7:PHRASE` substructure; the `g7:TYPE` substructure of the `KIND`'s superstructure should be used for that purpose instead.
+
+#### `KIND` (Kind) `g7:MARR-KIND`
+
+An enumerated value from set `g7:enumset-MARR-KIND` indicating what kind of marriage the superstructure was.
```
tychonievich, line 2011:
```suggestion
An enumerated value from set `g7:enumset-MARR-KIND` indicating what kind of marriage the superstructure describes.
```

tychonievich:
Discussed in steering committee. We like this draft, but plan to wait a bit before merging to allow others who have commented on these ideas in the past to weigh in

Norwegian-Sardines:
> I didn't include cohabitating/partnership/companionship marriage kinds because I thought some researchers might class those as being closer to a `NO MARR` than a `MARR`, but I'm not sure that was the right decision.

The introduction of the `NO` clause (non_event_structure) on events and attributes (no-date says it applies to attributes as well) throws a monkey wrench into a lot of logic. In this example, is a `NO MARR` saying that they never married (ever) or that you never found a marriage document confirming the marriage?  Historically, marriage documents may be non existent (either lost or never created), or the recording of the union took place many years (and children) later because the remoteness of the farm or lack of an active church for recording.

Personally for me a `NO` of any enum would require proof that an event/attribute did not occur `NO BAPM`, never baptized or no record of a baptism, not dead or no record of a death?  So for me `NO MARR` would need a qualification anyway saying they “cohabitated” if they lived together, or if they created a child but never lived together “affaire | liaison” which would still be implied if they used `MARR` and a `KIND` of affaire | liaison or cohabitated!

tychonievich:
> In this example, is a `NO MARR` saying that they never married (ever) or that you never found a marriage document confirming the marriage?

Per the spec, it means the former: 

> `1 NO MARR` means “no marriage occurred”

As with all factual claims, it would hopefully be provided with a source, such as an obituary stating that a partner in the FAM was unmarried. A source generated during their life would presumably use the date period to restrict it, as

```gedcom
0 FAM
1 NO MARR
2 DATE TO 3 JUN 1901
2 SOUR @S1@
```

where S1 is a document listing one of the partner's marital status as single never married.

My hesitation in including cohabitating as a MARR type is that the first few example records mentioning cohabitation I found in a more-or-less random search all seemed to me to be suitable sources of a `NO MARR`.`DATE TO ...` structure.

Norwegian-Sardines:
The making of a family unit has several components.  “Cohabitation” is a `NO MARR` when a child is involved and they continued to live as a family but never married, just like “affair | liaison” would also be a `NO MARR` if they had a child but never lived together as a family.

I have both of these conditions in my genealogy in both the 20th and 19th centuries!

tychonievich:
Discussed 2023-08-03
There have been no objections to this proposal, so we are merging it into next-minor. There have been possible proposed additions (MARR.KIND COHABITATED or the like) which may be revisited as a separate pull request.

PR #324, opened 2023-07-10
Age payloads are optional

### LINKED ISSUE #323: Optional AGE payload
The Age datatype includes the text

> Age payloads may also be omitted entirely if no suitable form is known but a substructure (such as a [PHRASE](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#PHRASE)) is desired.

but the ABNF for `Age` does not allow empty payloads.

See also #81 which raised a similar issue (now resolved) with the Date datatypes

tychonievich:
resolves #323


tychonievich:
@dthaler Should we add this test case to maximal.ged?

PR #344, opened 2023-08-31
recommend 1 FAMC per FAM and 1 CHIL per INDI

### LINKED ISSUE #339: Can an INDI have two FAMC to the same FAM?
In PR #274 we added to next-minor multiple PEDI for the same FAMC, which is also referenced in #338. But I now thing we should roll that change back as we can simple use several FAMC instead.

Two-FAMC version, keeping `PEDI` as `{0:1}` cardinality:

```gedcom
0 @F1@ FAM
1 CHIL @I1@
0 @I1@ INDI
1 FAMC @F1@
2 PEDI FOSTER
3 DATE FROM 1901 TO 24 JUN 1904
1 FAMC @F1@
2 PEDI ADOPTED
3 DATE FROM 24 JUN 1904
```

Two-PEDI version:

```gedcom
0 @F1@ FAM
1 CHIL @I1@
0 @I1@ INDI
1 FAMC @F1@
2 PEDI FOSTER
3 DATE FROM 1901 TO 24 JUN 1904
2 PEDI ADOPTED
3 DATE FROM 24 JUN 1904
```

Although it seems odd to have two FAMC pointing to the same FAM, I don't see any prohibition of it in the current spec; thus I think that making PEDI {0:M} would create ambiguity with two ways of encoding the same data.

tychonievich:
This is intended to resolve #339 (and should be copied to next-minor as well)

### DIFF
--- a/specification/gedcom-3-structures-1-organization.md
+++ b/specification/gedcom-3-structures-1-organization.md
@@ -224,6 +224,8 @@ the `INDI` record must use `FAMS` to point to the `FAM` record.
 If a `FAM` record uses `CHIL` to point to an `INDI` record,
 the `INDI` record must use a `FAMC` to point to the `FAM` record.
 
+An `INDI` record should not have multiple `FAMS` substructures pointing to the same `FAM`.
+
 A `FAM` record should not have multiple `CHIL` substructures pointing to the same `INDI`; doing so implies a nonsensical birth order.
 An `INDI` record may have multiple `FAMC` substructures pointing to the same `FAM`, but doing so is not recommended.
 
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-1-organization.md
```diff
@@ -224,6 +224,8 @@ the `INDI` record must use `FAMS` to point to the `FAM` record.
 If a `FAM` record uses `CHIL` to point to an `INDI` record,
 the `INDI` record must use a `FAMC` to point to the `FAM` record.
 
+A `FAM` record should not have multiple `CHIL` substructures pointing to the same `INDI`; doing so implies a nonsensical birth order.
+An `INDI` record may have multiple `FAMC` substructures pointing to the same `FAM`, but doing so is not recommended.
```
tychonievich, line 228:
```suggestion
An `INDI` record should not have multiple `FAMS` substructures pointing to the same `FAM`.

A `FAM` record should not have multiple `CHIL` substructures pointing to the same `INDI`; doing so implies a nonsensical birth order.
An `INDI` record may have multiple `FAMC` substructures pointing to the same `FAM`, but doing so is not recommended.
```

PR #349, opened 2023-09-27
Add ADOPTED to g7:enumset-NAME-TYPE

### LINKED ISSUE #335: Add ADOPTED to g7:enumset-NAME-TYPE
Since adoption is a concept, which is already used in several parts of the GEDCOM specification, it should also be available within the name type.

Therefore, ADOPTED should be added as an enum value to g7:enumset-NAME-TYPE

dthaler:
This only adds ADOPTED, which we had consensus on, without precluding the addition of other values in the future.

Fixes #335

### DIFF
--- a/specification/gedcom-3-structures-4-enumerations.md
+++ b/specification/gedcom-3-structures-4-enumerations.md
@@ -224,7 +224,7 @@ They are no longer used in that way, meaning their interpretation is subject to 
 
 | Value | Meaning                       |
 | ----- | :---------------------------- |
-|  ADOPTED | Name given at the time of adoption. |
+|  `ADOPTED` | Name given at the time of adoption. |
 | `AKA` | Also known as, alias, etc. |
 | `BIRTH` | Name given at or near birth. |
 | `IMMIGRANT` | Name assumed at the time of immigration. |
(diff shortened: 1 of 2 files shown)


specification/gedcom-3-structures-4-enumerations.md
```diff
@@ -224,6 +224,7 @@ They are no longer used in that way, meaning their interpretation is subject to
 
 | Value | Meaning                       |
 | ----- | :---------------------------- |
+|  ADOPTED | Name given at the time of adoption. |
```
tychonievich, line 227:
```suggestion
|  `ADOPTED` | Name given at the time of adoption. |
```

small formatting tweak

specification/gedcom-3-structures-4-enumerations.md
```diff
@@ -224,6 +224,7 @@ They are no longer used in that way, meaning their interpretation is subject to
 
 | Value | Meaning                       |
 | ----- | :---------------------------- |
+|  `ADOPTED` | Name given at the time of adoption. |
```
dthaler, line 227:
```suggestion
| `ADOPTED` | Name given at the time of adoption. |
```

dthaler:
Agreed during GEDCOM steering committee 9/28/2023

PR #350, opened 2023-09-27
Recommend that YAML files be available for extension structures

### LINKED ISSUE #330: SCHEMA.TAG ambiguity
https://github.com/FamilySearch/GEDCOM.io/pull/89 has an illustrative case:
```
1 SCHMA
2 TAG _PARTY http://example.com/party-participation
2 TAG _PARTY http://example.com/party
```

In the example above, the two URIs may differ in terms of what superstructures `_PARTY` can appear in.
However, there's no way for a GEDCOM processor to know which is the case without knowing the schema details.
Should there be a substructure under `TAG` that can list the superstructures? Or do we require the URI to resolve to a YAML file?  Or do we leave it unusable by generic tools?

dthaler:
Per discussion in #330:
> Discussed 2023-08-10
We think recommending that YAML files be available from registry or URL may be a simpler approach than adding content to the header.

Fixes #330


tychonievich (approved):

dthaler:
Approved text during GEDCOM steering committee 9/28/2023

PR #367, opened 2023-10-26
Clarify that FILE-TRAN has a FILE-like payload

### LINKED ISSUE #362: Separate datatype for file-path payloads
We should make it clear that `g7:FILE-TRAN` follows the same rules with respect to schemes, GEDZIPs, and so on as `g7:FILE`. This is only loosely implied in the current spec.

In the current spec, `g7:FILE-TRAN` is defined as

> A type of [TRAN](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#TRAN) for external media files. Each [g7:NOTE-TRAN](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#NOTE-TRAN) must have a [FORM](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FORM) substructure. See also [FILE](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FILE).
> 
> Note that [FILE.TRAN](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FILE-TRAN) refers to translation to a different digital format, not to translation to a different human language. Files that differ in the human language of their content should each be given their own [FILE](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FILE) structure.

Nowhere in that does it make it clear that all the rules for file paths that are listed under `g7:FILE` apply to the `g7:FILE-TRAN` too. Additionally, the GEDZIP section mentions `FILE` repeatedly but never mentions `g7:FILE-TRAN`.

I think the easiest way to fix this ambiguity is to add a file path datatype:

1. Add another datatype, maybe `g7:type-Path`.
2. Move the documentation for g7:FILE's payload to that datatype.
3. Update GEDZIP to refer to that datatype.
4. Change `g7:FILE` and `g7:FILE-TRAN` to have that datatype as their payload.

Adding a datatype is a bigger change than we usually do as a patch, but I think it is allowed because it doesn't change which files are valid or what they mean, only how we describe them.

If we don't want to add a datatype, an alternative option is to

1. Add text to `g7:FILE-TRAN` making it explicit that it has the same rules about its payload as `g7:FILE`.
2. Change everywhere GEDZIP refers to "`FILE`" to refer to "`g7:FILE` and `g7:FILE-TRAN`" instead.
3. Do one of the following
    1. State that these two structures are the only ones, and extensions should use them as substructures instead of adding a file path as their payload; or
    2. Add some non-datatype marker of "file path payload" in the YAML files

tychonievich:
One of two ways to resolve #362 (the other is #368).

- Document that FILE-TRAN payloads must follow the same formatting rules as FILE

- Document that GEDZIP must consider both FILE and FILE-TRAN, not just FILE

This makes minimal changes to the spec (and, like the other proposal, no changes to GEDCOM files). It requires applicatiosn to treat the payloads of these two structure types specially and does not support extensions adding additional file-path-payload structure types.

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1505,7 +1505,8 @@ if the resulting text is different from the text created by the HTML-to-text con
 #### `TRAN` (Translation) `g7:FILE-TRAN`
 
 A type of `TRAN` for external media files.
-Each `g7:NOTE-TRAN` must have a `FORM` substructure.
+Each `g7:FILE-TRAN` must have a `FORM` substructure.
+`g7:FILE-TRAN` payloads follow the same rules as `g7:FILE` payloads.
 See also `FILE`.
 
 :::example
--- a/specification/gedcom-4-gedzip.md
+++ b/specification/gedcom-4-gedzip.md
@@ -11,7 +11,7 @@ Each GEDZIP file contains the following entries:
 
 - An entry with name `gedcom.ged` containing a data stream.
 
-- An entry for each *local file* `FILE` structure in `gedcom.ged`, with the same zip *file name* as the corresponding `FILE` payload.
+- An entry for each *local file* `FILE` and `FILE-TRAN` structure in `gedcom.ged`, with the same zip *file name* as the corresponding `FILE` payload.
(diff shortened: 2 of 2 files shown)


dthaler (commented):
I like the other alternative better, as it is more amenable to automated tooling.

tychonievich:
Steering committee chose #368 instead

PR #368, opened 2023-10-26
Change FILE and FILE-TRAN payloads from Special to FilePath

### LINKED ISSUE #362: Separate datatype for file-path payloads

tychonievich:
This adds a file path datatype, one of two options to resolve #362 (the other is #367). It adds a new datatype to the spec, but makes no change to the resulting GEDCOM files. The new datatype may make automated tooling and extensions with file paths easier to define.

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -638,7 +638,7 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
 #### `FILE` (File reference) `g7:FILE`
 
 A reference to an external file.
-See the [File Path datatype](#file-path) for more.
+See the [File Path datatype](#file-path) for more details.
 
 #### `FORM` (Format) `g7:FORM`
 
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -638,32 +638,7 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
 #### `FILE` (File reference) `g7:FILE`
 
 A reference to an external file.
-Syntactically, the payload is a URI reference as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), or a valid URL string as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).
-That is, it can be an absolute or relative URL, optionally with a fragment string.
-However, only the following URL types are used:
-
-- A URL with scheme `ftp`, `http`, or `https` refers to a **web-accessible file**.
-
-- A URL with scheme `file` refers to a **machine-local file** as defined by [RFC 8089](https://www.rfc-editor.org/info/rfc8089). Machine-local files must not be used in [FamilySearch GEDZIP](#gedzip) nor when sharing datasets on the web or with unknown parties, but may be used for close collaboration between parties with known similar file structures.
-
-- A URI reference with all of the following:
-    - no scheme
-    - not beginning with `/` (U+002F)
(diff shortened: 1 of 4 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -638,32 +638,7 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
...
-    It is recommended that local files use the directory prefix `media/`, but doing so is not required.
-
-    For compatibility with [GEDZIP](#gedzip) and related formats, it is recommended that the following `FILE` payloads not be used:
-    
-    - `gedcom.ged`
-    - `MANIFEST.MF`
-    - any URL beginning `META-INF/`
-
-The meaning of a `FILE` payload with any format not listed above is not defined by this version of the specification, but may be defined in a subsequent version.
+See the [File Path datatype](#file-path) for more.
```
dthaler, line 641:
```suggestion
See the [File Path datatype](#file-path) for more details.
```
minor grammar suggestion

dthaler (approved):

specification/gedcom-3-structures-3-meaning.md
tychonievich: The sentence "See [...] for more." appears 50 times in the current spec; "for more details" does not appear at all. If we want to change these to "for more details" I suggest we do so as a single PR for all 50.

PR #373, opened 2023-11-16
Source stewardship support

dthaler:
Added DATE under repository citation

Per https://github.com/dthaler/gedcom-citations/pull/6 extension

### DIFF
--- a/specification/gedcom-3-structures-1-organization.md
+++ b/specification/gedcom-3-structures-1-organization.md
@@ -1229,6 +1229,7 @@ n REPO @<XREF:REPO>@                       {1:1}  g7:REPO
   +1 CALN <Special>                        {0:M}  g7:CALN
      +2 MEDI <Enum>                        {0:1}  g7:MEDI
         +3 PHRASE <Text>                   {0:1}  g7:PHRASE
+  +1 DATE <DateValue>                      {0:1}  g7:DATE
 ```
 
 This structure is used within a source record to point to a name and address record of the holder of the source document.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-1-organization.md
```diff
@@ -1237,3 +1238,7 @@ Formal and informal repository name and addresses are stored in the
 More formal repositories, such as the Family History Library, should show a call number of the source at that repository.
 The call number of that source should be recorded using a `CALN` substructure.
 
+A `DATE` substructure can be used to denote the date when the source was known to be held by the
+specific repository.  For example, if a Family Bible was passed from one person to another over time, a
```
tychonievich, line 1242:
```suggestion
specific holder.  For example, if a Family Bible was passed from one person to another over time, a
```

The paragraph above uses "holder" as a more general term than "repository"

specification/gedcom-3-structures-1-organization.md
```diff
@@ -1229,6 +1229,7 @@ n REPO @<XREF:REPO>@                       {1:1}  g7:REPO
   +1 CALN <Special>                        {0:M}  g7:CALN
      +2 MEDI <Enum>                        {0:1}  g7:MEDI
         +3 PHRASE <Text>                   {0:1}  g7:PHRASE
+  +1 DATE <DateValue>                      {0:1}  g7:DATE
```
tychonievich, line 1232:
```suggestion
  +1 DATE <DateValue>                      {0:M}  g7:DATE
```

Couldn't we have several dates as it was transferred to and from the same holder?

dthaler: Don't have a strong case for M currently, but changing to M in future would be non-breaking whereas changing from M to 1 would be a breaking change, so discussed staying with 1 for now.

Issue #364, opened 2023-10-19
Adding substructures in next-minor requires new URI


tychonievich:
I'm not sure if we've done this yet, but if any structure is modified in 7.1 to have a substructure it does not in 7.0 then it needs a new URI in 7.1

tychonievich:
The current next-minor branch has the following possibly-URI-impacting changes:

## Enumset values:
Should the sets get a new URI to reflect the changed members? Or the members get a new URI to reflect the changed set? Or both?

|set|new value|notes|
|-|-|-|
|g7:enumset-NAME-TYPE|g7:enum-ADOPTED|with slightly different definition than old enum-ADOPTED; perhaps should rename within this set?|
|g7:enumset-MEDI|g7:enum-DIGITAL
|g7:enumset-MEDI|g7:enum-ONLINE


## Enumsets of structure:
Should the structures get a new URI to reflect the changed set?

|Structure|Old|New|Notes|
|-|-|-|-|
|g7:NO|g7:enumset-EVEN|g7:enumset-NO|I'm not seeing a definition of this set in next-minor|

## Substructures:
The superstructures should probably get a new URI, to reflect the added substructure. My gut is the substructures do not get a new URI, but they could if we want.

|superstructure|new substructure|
|-|-|
|g7:record-SUBM|g7:SUBM-ALIA|
|g7:PEDI|g7:DATE|
|(all event/attr)| g7:EXID|
|g7:FAMC| g7:EXID|
|g7:INDI-NAME| g7:EXID|
|g7:NO| g7:EXID|
|g7:NOTE| g7:EXID|
|g7:SEX| g7:EXID|
|g7:SOUR| g7:EXID|
|g7:BURI|g7:BURI-KIND|
|g7:MARR|g7:MARR-KIND|
|g7:INDI-NAME|g7:LANG|
|g7:INDI-NAME|g7:NO|
|g7:TIME|g7:PHRASE|
|g7:NOTE|g7:REFN|
|g7:NO|g7:REFN|
|g7:INDI-NAME|g7:RESN|
|g7:NO|g7:RESN|
|g7:record-REPO|g7:RESN|
|g7:record-SNOTE|g7:RESN|
|g7:record-SOUR|g7:RESN|
|g7:record-SUBM|g7:RESN|
|g7:SOUR|g7:RESN|
|g7:DATE-exact|g7:TIME-exact|new URI for substructure; was TIME|
[...]

PR #374, opened 2023-11-16
Create a v7.1 URI prefix

dthaler:
For issue #364

A large number of URIs changed just due to `NOTE_STRUCTURE` and `SOURCE_CITATION` allowing a `RESN` in 7.1.

### DIFF
--- a/specification/gedcom-0-introduction.md
+++ b/specification/gedcom-0-introduction.md
@@ -147,6 +147,7 @@ is shorthand for a URI beginning with the corresponding URI prefix
 | Short Prefix | URI Prefix                          |
 |:-------------|:------------------------------------|
 | `g7`         | `https://gedcom.io/terms/v7/`       |
+| `g7.1`       | `https://gedcom.io/terms/v7.1/`     |
 | `xsd`        | `http://www.w3.org/2001/XMLSchema#` |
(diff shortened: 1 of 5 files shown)


specification/gedcom-0-introduction.md
```diff
@@ -147,6 +147,7 @@ is shorthand for a URI beginning with the corresponding URI prefix
 | Short Prefix | URI Prefix                          |
 |:-------------|:------------------------------------|
 | `g7`         | `https://gedcom.io/terms/v7/`       |
+| `g7.1`       | `https://gedcom.io/terms/v7.1/`     |
```
tychonievich, line 150: Should we use `71` or `7-1` or some other prefix that does not have a period so that the only period in URLs is used for file endings? Doesn't matter for most modern applicants, I suppose, but might for some older apps?

specification/gedcom-3-structures-1-organization.md
```diff
@@ -236,15 +236,15 @@ An `INDI` record may have multiple `FAMC` substructures pointing to the same `FA
 n @XREF:INDI@ INDI                         {1:1}  g7:record-INDI
```
dthaler, line 236: and propagate it all the way up

dthaler:
Abandoning for now, will recreate after new v7.1 branch is created

PR #398, opened 2023-12-06
Allow non-local file URLs in GEDZIP

### LINKED ISSUE #397: Non-local file URIs in GEDZIP
[The GEDCOM 7 spec ](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FILE) says:

> A URL with scheme file refers to a **machine-local file** as defined by [RFC 8089](https://www.rfc-editor.org/info/rfc8089). Machine-local files must not be used in [FamilySearch GEDZIP](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#gedzip) nor when sharing datasets on the web or with unknown parties, but may be used for close collaboration between parties with known similar file structures.

There are a couple of problems with the above text:

1. The term "machine-local file" is never used by RFC 8089.  It does define **local file** though in [Section 1.1](https://www.rfc-editor.org/rfc/rfc8089.html#section-1.1):

> The term "local file URI" is used to describe file URIs that have no
> "authority" component or where the authority is the special string
> "localhost" or a fully qualified domain name that resolves to the
> machine from which the URI is being interpreted ([Section 2](https://www.rfc-editor.org/rfc/rfc8089.html#section-2)).

2. The current spec text implies that a URL with scheme file refers to a **local file**, but this is incorrect.  RFC 8089 defines file: for both **local files** and **non-local files**.  For example, [Appendix B](https://www.rfc-editor.org/rfc/rfc8089.html#appendix-B) has:

> Non-local files:
>
>   o  A non-local file with an explicit authority.  For example:
>
>      *  "file://host.example.com/path/to/file"

[Appendix E.3.2](https://www.rfc-editor.org/rfc/rfc8089.html#section-3) has:

> Non-local files:
> 
>    o  The representation of a non-local file with an empty authority and
>       a complete (transformed) UNC string in the path.  For example:
> 
>       *  "file:////host.example.com/path/to/file"
> 
>    o  As above, with an extra slash between the empty authority and the
>       transformed UNC string, as per the syntax defined in [[RFC1738](https://www.rfc-editor.org/rfc/rfc1738)].
>       For example:
> 
>       *  "file://///host.example.com/path/to/file"
> 
>       This representation is notably used by the Firefox web browser.
>       See Bugzilla#107540 [[Bug107540](https://www.rfc-editor.org/rfc/rfc8089.html#ref-Bug107540)].

and [Section 3](https://www.rfc-editor.org/rfc/rfc8089.html#section-3) states:

>    This specification neither defines nor forbids any set of operations
>    that might be performed on a file identified by a non-local file URI.

3. The current spec text does not say what to do with GEDZIP when the GEDCOM file contains a non-local file URI.  It only says what to do with local files and ftp, http, and https URLs.

* Option 1: Treat non-local file URIs the same as ftp.  Allowed in FamilySearch GEDZIP.
* Option 2: Leave out the FILE structure.   This loses information, causing the GEDZIP to not contain all info that the GEDCOM has.
* Option 3: Require downloading the non-local file so that it becomes a local file, and then include it as a local file.  This can result in the GEDZIP occupying _more_ space than the GEDCOM, defeating the benefits of GEDZIP.

dthaler:
This change permits non-local file URLs to be used in GEDZIP.

Fixes #397

### DIFF
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -339,7 +339,7 @@ Version 7.0 only supports the following URLs:
 
 - A URL with scheme `ftp`, `http`, or `https` refers to a **web-accessible file**.
 
-- A URL with scheme `file` refers to a **machine-local file** as defined by [RFC 8089](https://www.rfc-editor.org/info/rfc8089). Machine-local files must not be used in [FamilySearch GEDZIP](#gedzip) nor when sharing datasets on the web or with unknown parties, but may be used for close collaboration between parties with known similar file structures.
+- A URL with scheme `file` refers to either a **local file** or a **non-local file**, as defined by [RFC 8089](https://www.rfc-editor.org/info/rfc8089). Local file URLs must not be used in [FamilySearch GEDZIP](#gedzip) nor when sharing datasets on the web or with unknown parties, but may be used for close collaboration between parties with known similar file structures.
 
 - A URI reference with all of the following:
     - no scheme


dthaler:
GEDCOM Steering Commitee discussion:
Spec was ambiguous, could be read as either no file URIs in gedzip or no local file URIs in gedzip.
Disallowing all file URIs in gedzip has the problems in issue #397 and may break things.

Issue #401, opened 2023-12-07
Requirement re "shared on the web" can't be a requirement


tychonievich:
The current spec includes (emphasis added)

> Machine-local files must not be used in [FamilySearch GEDZIP](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#gedzip) **nor when sharing datasets on the web or with unknown parties**, but may be used for close collaboration between parties with known similar file structures.

This makes a requirement on the file based on an unknowable constraint of how the file will be used, and thus a requirement that cannot be satisfied in general. It should be weakened to a recommendation instead.

dthaler:
@albertemmerich is this ok to close now that #406 is merged?

dthaler:
Discussion 12/28/2023: committee believes this issue is now addressed.  @albertemmerich please reopen if you disagree.  Thanks!

PR #406, opened 2023-12-13
Change local file restriction to a recommendation

tychonievich:
This change does technically remove a requirement and thus expand the set of files that the specification describes. However, the requirement it removes was ill-defined and unenforceable, as noted in issue #401, which is why this PR is being submitted as a patch instead of a new minor version.

Resolves issue #401


dthaler (approved):

lynchrs3 (approved):

dthaler:
Discussion 12/21/2023: Looks ok, so merging.  @albertemmerich is this sufficient to close #401?

PR #407, opened 2023-12-13
Clarify the extra column in enumset-ord-STAT

### LINKED ISSUE #400: 7.0.10's change to enumset-ord-STAT needs to be clarified
7.0.10 added an "applies to" column to the ord-STAT enumset, but did not define what it means. Because this was a patch, not a major version, it can at most be interpreted as a recommendation, not a restriction, but that is not clear in the specification text and has been interpreted as a restriction in several conversations, including the conversation that motivated PR #399

tychonievich:
Resolve #400

`g7:enumset-ord-STAT` has "Applies to" and "Status" columns that are recommendations, not requirements, but it fails to make that distinction clear. This change clarifies their meaning.


dthaler (approved):

lynchrs3 (approved):

dthaler:
Approved during GEDCOM Steering Committee 12/21/2023

PR #409, opened 2023-12-17
Update gedstruct to make cross reference identifiers optional

### LINKED ISSUE #408: Are cross-reference identifiers optional
Section 1.3 says:

> Each record to which other structures point must have a cross-reference identifier. A record to which no structures point may have a cross-reference identifier, but does not need to have one. A substructure or pseudo-structure must not have a cross-reference identifier.

However the gedstruct throughout the document contradicts this currently, since it has lines like this for each record:

```gedstruct
n @XREF:INDI@ INDI                  {1:1}  g7:record-INDI
```

Where the cross-reference identifier is not denoted as null being permitted.  To be consistent with the paragraph quoted, the gedstruct would need to be

```gedstruct
n [@XREF:INDI@|<NULL>] INDI         {1:1}  g7:record-INDI
```

As such, validators that check for conformance against the gedstruct disallow files without the cross-reference identifier.

The ambiguity is, between the paragraph or the gedstruct, which is correct and which is incorrect.

dthaler:
Fixes #408

### DIFF
--- a/specification/gedcom-3-structures-1-organization.md
+++ b/specification/gedcom-3-structures-1-organization.md
@@ -175,7 +175,7 @@ A few substructures of note:
 #### `FAMILY_RECORD` :=
 
 ```gedstruct
-n @XREF:FAM@ FAM                           {1:1}  g7:record-FAM
+n [@XREF:FAM@|<NULL>] FAM                  {1:1}  g7:record-FAM
   +1 RESN <List:Enum>                      {0:1}  g7:RESN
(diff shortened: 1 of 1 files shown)


tychonievich (changes requested):
This is changing the "cross-reference template" part of the gedstruct syntax.

I disagree with this change (see [my comment in #408](https://github.com/FamilySearch/GEDCOM/issues/408#issuecomment-1859342457) for more), which makes the syntax more complicated and less compatible with the syntax used in the various 5.x versions.

If we do make this change we must also change the definition of the syntax in section 3.1, as this edit violates the current definition.

Issue #97, opened 2021-11-23
Add SOUR.PHRASE in future and _PHRASE now


tychonievich:
Any structure that can have a @VOID@ pointer might need a PHRASE. We should add this in a future version and document the user of _PHRASE for this purpose now.

_Originally posted by @tychonievich in https://github.com/FamilySearch/GEDCOM/discussions/89#discussioncomment-1690687_

tychonievich:
Also add this in the migration guide

tychonievich:
Discussed 2021-12-14
Plan is to present both options (NOTE and _PHRASE), recommend _PHRASE, and state that we plan to resolve this in a future version of the specification.

dthaler:
I've looked into this further since we last met, to see what my code does.   My code follows the recommendation in https://gedcom.io/migrate/#non-pointer-sour-substructures which says that any 5.5.1 `SOURCE_CITATION`s with no `SOURCE_RECORD` should be converted to `SOURCE_CITATION`s with `SOURCE_RECORD`s, in which case this problem does not arise.   Upon reflection, I can see a possible argument to support it without `SOURCE_RECORD`s (though my first preference would be do require it) if there are systems that do not support source records at all and don't want to for GEDCOM 7.  I hope not.   GEDCOM 5.5.1 already stated that using source records was "preferred" so one could consider the lack of them arguably deprecated 20 years ago.

But if it is required then I think it is like `NOTE` where we created `NOTE` and `SNOTE` as separate tags, rather than using `NOTE @VOID@` with a `PHRASE` substructure.  Since GEDCOM 7 already uses SOUR with an xref, this approach would entail a separate tag for the void case.

So some options:

Option 1) Do not standardize source citations without a source record.

Option 2) Make SOUR follow the NOTE precedent by creating a new tag in a future version for a source citation without a source record.

Option 3) Make NOTE and SOUR work differently even though they solve the same problem.   NOTE/SNOTE use different tags whereas SOUR adds a new PHRASE substructure.

Those are listed in order of my preference.   I don't like option 3 any more because it seems to add new inconsistency that doesn't exist now.

If option 2 or 3 are selected, then we still have multiple ways to convert a 5.5.1 citation to 7, and we would need to decide what to recommend in the migration guide when converting from a source citation without a source record:

Option A) Recommend (as 5.5.1 prefers, and as the current migration guide says to do) using a source record in GEDCOM 7.

[...]

PR #417, opened 2024-01-07
Deprecate using stdTag in non-standard ways

dthaler:
As discussed in issue #97, there's no way in YAML to express using a stdTag for an extension-defined substructure that doesn't use a standard type URI.  As such, we said uses like `_LOC`.`NAME` that permit `DATE` as a substructure should be deprecated.

### DIFF
--- a/specification/gedcom-1-hierarchical-container-format.md
+++ b/specification/gedcom-1-hierarchical-container-format.md
@@ -319,7 +319,7 @@ Extensions cannot change existing meanings, cardinalities, or calendars.
 A **tagged extension structure** is a structure whose tag matches production `extTag`. Tagged extension structures may appear as records or substructures of any other structure. Their meaning is defined by their tag, as is discussed more fully in the section [Extension Tags].
 
 Any substructure of a tagged extension structure that uses a tag matching `stdTag` is an **extension-defined substructure**.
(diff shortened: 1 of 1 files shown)


specification/gedcom-1-hierarchical-container-format.md
```diff
@@ -319,7 +319,7 @@ Extensions cannot change existing meanings, cardinalities, or calendars.
 A **tagged extension structure** is a structure whose tag matches production `extTag`. Tagged extension structures may appear as records or substructures of any other structure. Their meaning is defined by their tag, as is discussed more fully in the section [Extension Tags].
 
 Any substructure of a tagged extension structure that uses a tag matching `stdTag` is an **extension-defined substructure**.
-Substructures of an extension-defined substructure that uses a tag matching `stdTag` are also extension-defined substructures.
+Substructures of an extension-defined substructure that uses a tag matching `stdTag` are also extension-defined substructures, but this specification deprecates using a `stdTag` with a different definition than a standard type with that tag.
```
dthaler, line 322:
```suggestion
Substructures of an extension-defined substructure that uses a tag matching `stdTag` are also extension-defined substructures, but this specification deprecates using a `stdTag` with a definition that does not match any standard type with that tag.
```

dthaler:
Agreed during GEDCOM meeting 1/18/2024

Issue #420, opened 2024-01-18
AGE payload issues


dthaler:
Are the following AGE payloads legal?

```
AGE 8w 30d
AGE 1y 400d
AGE 1y 30m
AGE 1m 40d
```

Section 2.4 says about dates:
> The allowable [day](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date)s, [month](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date)s, [year](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date)s, and [epoch](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date)s are determined by the [calendar](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date). All known calendars restrict [day](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#date) to be between 1 and a month-specific maximum. The largest known maximum is 36, and most months in most calendars have a lower maximum.

But nothing says that AGE payloads have any similar restrictions, implying that all four examples are permitted, and are not even discouraged.

A closely related problem is that say someone has events in two different calendars, say both `FRENCH_R` and `GREGORIAN`, then their age in terms of months and extra days can be _different_ in each calendar.  The `AGE` tag has no calendar permitted in the payload.  One might infer it from the `DATE` of the event, if there is one, but `AGE` is permitted in cases with no `DATE`.  I would expect most applications to store it in GEDCOM exactly as it was given in the original record, but that is not interoperable if other programs use it to calculate the date but interpret it in the wrong calendar.

Some possible actions:
1. We could possibly add a Note saying the uses up top are not recommended (I think we can't disallow them without a breaking change) and applications should use the simplest form.  Rationale to NOT add such a note might be to allow recording ages as they literally appear in a record.  But a Note that applications must be prepared to see them might be helpful.
2. Recommend that when `AGE` is used, there should be a `DATE` with it, where a `CAL` payload can be used to disambiguate which calendar.  Otherwise, some programs might calculate incorrect dates based on using a calendar other than the correct one for the event.

dthaler:
Discussion in GEDCOM Steering Committee1/18/2024:
* We have no way to store the calendar of an age
* A DATE and AGE might both be in an original record and so don't necessarily match
* Could be discussed in a FAQ
* Having a Note in the spec is ok if wording is right, to promote more interoperability
* Lines at top are ok to be in test files
[...]

PR #432, opened 2024-02-01
Add note about oddly formed age payloads

dthaler:
Add note in the GEDCOM spec about oddly formed age payloads being permitted.

Addresses part of issue #420 

I'm still not entirely happy with this result, especially since `AGE.PHRASE` can always be used to store the original.
The reason I'm not happy is that it means an application must be able to store age payloads with out-of-normal-range values like age in months that requires more than (say) 1 byte to store.  This can cause headaches with application data model design.

### DIFF
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -225,6 +225,12 @@ Because numbers are rounded down, `>` effectively includes its endpoint; that is
 
 Different cultures count ages differently. Some increment years on the anniversary of birth and others at particular seasons. Some round to the nearest year, others round years down, others round years up. Because users may be unaware of these traditions or may fail to convert them to the round-down convention, errors in age of up to a year are common.
 
+:::note
+Because age payloads are intended to allow recording the age as it was recorded in records that could contain errors,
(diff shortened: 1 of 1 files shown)


specification/gedcom-2-data-types.md
```diff
@@ -225,6 +225,12 @@ Because numbers are rounded down, `>` effectively includes its endpoint; that is
 
 Different cultures count ages differently. Some increment years on the anniversary of birth and others at particular seasons. Some round to the nearest year, others round years down, others round years up. Because users may be unaware of these traditions or may fail to convert them to the round-down convention, errors in age of up to a year are common.
 
+:::note
+Because age payloads are intended to allow recording the age as it was recorded in records that could contain errors,
+odd ages such as `8w 30d`, `1y 400d`, `1y 30m`, etc. are permitted.  Some applications might convert these to more
+standard forms and use a `PHRASE` substructure to hold the original form.
```
tychonievich, line 231:
```suggestion
standard forms; if so, it is recommended that they use a `PHRASE` substructure to hold the original form.
```

Change suggested to make it not seem like applications who standardize always use `PHRASE`

Issue #418, opened 2024-01-12
Machine readable schema - request


augean:
XML and JSON schemas are usually machine-readable
Requesting the same for GEDCOM 7

Please see the attached files, for an example of a machine-readable schema that I created for GEDCOM  5.1.1
This allows me to create new GEDCOM files easily,  and is very easy for tools to interact with

Please could we have a machine-readable GEDCOM  7 schema     (please use the attached files as an example)
I single file (or even multiple files),  which allows us to easily parse the GEDCOM  
[ged.5.1.1.txt](https://github.com/FamilySearch/GEDCOM/files/13909995/ged.5.1.1.txt)
[PrimSection.txt](https://github.com/FamilySearch/GEDCOM/files/13909996/PrimSection.txt)
structure

thanks

tychonievich:
7.0 has a machine-readable schema.

The line syntax is defined in the spec using ABNF, which is automatically extracted as [grammar.abnf](https://github.com/FamilySearch/GEDCOM/blob/main/extracted-files/grammar.abnf). At least one of the public [gedcom parsers](https://gedcom.io/tools/#v70-parsers) uses this grammar to parse lines.

The structure hierarchy is defined in the spec using a machine-readable variant of the metasyntax created for 5.0, which is automatically extracted as [grammar.gedstruct](https://github.com/FamilySearch/GEDCOM/blob/main/extracted-files/grammar.gedstruct). The structure hierarchy is also converted to a different machine-readable form as a set of YAML files hosted in several places including the URI of each structure type (e.g. <https://gedcom.io/terms/v7/ABBR>) and in a separate repository of both standard and extension structures ([GEDCOM-registries](https://github.com/FamilySearch/GEDCOM-registries)). Multiple public [gedcom parsers and development aids](https://gedcom.io/tools/#v70-parsers) use one or both of these to parse and validate structure hierarchies.

These machine-parseable formats are not perfect (for example, we lack a machine-parseable way of marking something as deprecated) and we'd welcome suggestions in how to improve them.  I did not look at your attached files closely enough to know if you have features the standard currently lacks.

dthaler:
Discussion in GEDCOM Steering Committee 1/18/2024:
We have machine readable schema.
We have machine readable positive test cases in the GEDCOM.io repository.
We currently don't have machine readable negative test cases, such as appear in [PrimSection.txt](https://github.com/FamilySearch/GEDCOM/files/13909996/PrimSection.txt)
[...]

PR #437, opened 2024-02-22
Add language tag and media type ABNF to grammar.abnf

dthaler:
This doesn't change the GEDCOM spec, only the grammar.abnf extracted file, to include ABNF extracted/derived from the RFCs referenced in the GEDCOM spec for language tags and media types.

This addresses the second bullet in the action item list at the bottom of issue #418

I filed issue #438 while working on this, but this PR could be merged without addressing that issue.


tychonievich (approved):

dthaler:
Agreed per meeting 22 FEB 2024

PR #445, opened 2024-03-07
remove reference to defunct next-major

tychonievich:


dthaler:
Approved during meeting March 7, 2024

PR #459, opened 2024-05-01
Add Separation event

cdhorn:
Add Separation to the family events table.


attribute-event-requests/FAM-event.md
```diff
@@ -85,3 +86,33 @@ In [GEDCOM X](https://github.com/FamilySearch/gedcomx/blob/master/specifications
...
+### Absence
+
+There is no other structure that covers this event, as it can apply to couples that cohabitate and are not married as well as to those who are married in some manner. And for those who are married, some couples may separate and never actually file for or go through with a formal divorce.
+
+### Used
+
+- Part of the [GEDCOM X specification](https://github.com/FamilySearch/gedcomx/blob/master/specifications/fact-types-specification.md) with URI `http://gedcomx.org/Separation`
+
+- Used by the [FamilySearch API](https://www.familysearch.org/developers/docs/guides/facts) via GEDCOM X.
+
```
tychonievich, line 113:
```suggestion
```

This event type is not documented in the FamilySearch API (at least not at the link provided). To the best of my knowledge, it is also not used by any FamilySearch application (I'm not a FamilySearch employee, but so far as I know they only use their API, not GEDCOM-X, in their products)

tychonievich (changes requested):
Thanks for adding this! Except for one suggested change about the difference between GEDCOM-X and the FamilySearch API, I think this looks good!

attribute-event-requests/FAM-event.md
cdhorn: Oops, yes you are correct I should have done due diligence. Fixed.

attribute-event-requests/FAM-event.md
```diff
@@ -85,3 +86,31 @@ In [GEDCOM X](https://github.com/FamilySearch/gedcomx/blob/master/specifications
...
+*proposed description missing*
+
+*In [GEDCOM X](https://github.com/FamilySearch/gedcomx/blob/master/specifications/fact-types-specification.md) as* "A fact of a couple's separation."
+
+### Value
+
+Found in the following historical records:
+
+- (records not yet identified)
```
dthaler, line 102: Question: Some records may record "separated" as a status rather than an event.  E.g., on 1 JUN 1980 the person was listed as "separated", which doesn't mean the separation happened that day.   Would that be recorded as "BEF 1 JUN 1980" if this were an event?  I'm comparing "Separation" with the "Lived Together" entry which is in FAM-attribute.md rather than FAM-event.md.  To me it seems closer to being like that concept so maybe they should be in the same .md file?

tychonievich (approved):

attribute-event-requests/FAM-event.md
dthaler: https://github.com/FamilySearch/GEDCOM/blob/main/attribute-event-requests/FAM-attribute.md#lived-together is where "Lived Together" is currently tracked.  The GEDCOM Steering Committee discussed this today, and we recommend taking this same text (which looks fine, thanks!) and moving it to FAM-attribute.md next to Lived Together.

cdhorn: @dthaler I would ask the committe to reconsider this. From what I have seen third party programs treat this as an event and not an attribute.

In my mind events are best thought of as experiences, something a subject either actively or passively participated in or underwent either alone or with others. And attributes are best thought of as characteristics or properties that describe something about a subject. Both are categories or types of facts, with the two others being identity and relationship.

I may have missed it but I've thought the lack of a clear up front definition or explanation to distinguish between these two categories or types is a shortcoming in the specification.

I've always thought the word attribute was a poor choice and think the [GenTech](https://xml.coverpages.org/GENTECH-DataModelV11.pdf) data model made things clearer by calling them characteristics instead.

cdhorn: @dthaler in case it is not clear from my comments above, I think "Separated" would be a characteristic that describes the state of the relationship between the couple over some interval in time while "Separation" is an experience that the couple went through.

The family record has always been problematic as it is "overloaded" and used for too many things. It serves as a limited proxy for the missing group concept, is used to describe the relationship between a couple, between the children and the parents, as a proxy for events shared between the couple or possibly the whole group, and attributes might apply to the relationship between the couple or the group as a whole as well.

Norwegian-Sardines: It has been a lot of years since I designed databases but one thing is true about GEDCOM, it is not a very well defined database.  

Yes, the Family_Record is overloaded, personally I don’t think it is even a candidate to be a record type at all!  Is “separated” a state of the `MARR` tag rather than a separate attribute/characteristic of the family?  

If we dive down into what is an attribute/characteristic of a family vs what is an event we need to do that for all such things within family and individual!  And don’t get me started on shared events between individuals vs what is a true attribute of an individual!

dthaler:
Discussion in GEDCOM Steering Committee meeting 9 MAY 2024:
* We are merging this based on the statement that existing programs may be doing this already.
* We will separately track the request to have a Separated attribute.

cdhorn:
@dthaler looked like you merged main into my branch and not the PR into main so just force pushed my branch again to reset it.

PR #480, opened 2024-06-13
Add the meaning of WWW

### LINKED ISSUE #476: Add explanation of `WWW` structure
In the specification, `g7:WWW` is defined only as having a URL payload, with no indication of what it means in context. The [specification text](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#WWW) is:

> A URL or other locator for a World Wide Web page, as defined by any relevant standard such as [whatwg/url](https://url.spec.whatwg.org/), [RFC 3986](https://www.rfc-editor.org/info/rfc3986), [RFC 3987](https://www.rfc-editor.org/info/rfc3987), and so forth.
>
> If an invalid or no longer existing web address is present upon import, it should be preserved as-is on export.

We should specify what it means. Because it has no specified meaning in 7.0 (or 5.5.1, which has even less text defining it) , we can't limit its meaning in a patch or minor version, but we can add recommendations and notes about how to handle it.

tychonievich:
Previously (in both 5.5.1 and 7.0.0–7.0.14) `WWW` was defined only by the type of is payload. This is an attempt to fix that without invalidating any existing files.

Resolves #476

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1642,12 +1642,19 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
 
 #### `WWW` (Web address) `g7:WWW`
 
-A URL or other locator for a World Wide Web page,
+A URL or other locator for a World Wide Web page of the subject of the superstructure,
 as defined by any relevant standard
 such as [whatwg/url](https://url.spec.whatwg.org/),
 [RFC 3986](https://www.rfc-editor.org/info/rfc3986),
 [RFC 3987](https://www.rfc-editor.org/info/rfc3987),
 and so forth.
+
+Like other substructures, the `WWW` structure provides details about the subject of its superstructure, not about some other subject.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -1642,12 +1642,19 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
...
 
-A URL or other locator for a World Wide Web page,
+A URL or other locator for a World Wide Web page of the subject of the superstructure,
 as defined by any relevant standard
 such as [whatwg/url](https://url.spec.whatwg.org/),
 [RFC 3986](https://www.rfc-editor.org/info/rfc3986),
 [RFC 3987](https://www.rfc-editor.org/info/rfc3987),
 and so forth.
 
+Like other substructures, the `WWW` structure provides details about the subject of its superstructure, not about some other subject.
```
dthaler, line 1652:
```suggestion
Like other substructures, the `WWW` structure provides details about the subject of its superstructure.
```

PR #482, opened 2024-06-14
Clarify nickname

### LINKED ISSUE #473: Add something for Dutch "Roepnaam" (CALLNAME)
As explained here: [https://www.webtrees.net/index.php/forum/2-open-discussion/37395-gedcom-7-personal-name-structure](url) by JustCarmen, Dutch people have trouble entering "Roepnaam".
That certainly is not a Nickname, but indeed a "Callname", so a name by which someone is known, is called so to speak in daily life.

I think "Roepnaam" is only used in the Netherlands, but I am not sure of that.

As stated in that link, and I will use myself as example now, at birth I was given the Firstnames "Catharina, Johanna, Elisabeth", and together with that my parents gave me the "Roepnaam": "Tineke".
So everyone knows me as Tineke and those 3 "Firstnames" are only used on official records.

So GIVN is "Catharina, Johanna, Elisabeth", NOT Tineke.
But Tineke is NOT my Nickname, my Nickname is Mother10.

As JustCarmen said in that link, maybe there should be a CALL for these cases.

Or maybe another NameType. In that case it might be possible to have that in the same minor as the other NameTypes.

Hope I made myself clear.

tychonievich:
Add additional clarification to nickname, explaining the word's meaning in English which is not shared by several European countries. See [this comment](https://github.com/FamilySearch/GEDCOM/issues/473#issuecomment-2168463707) and the rest of issue #473 for more on why this clarification is needed.

Although #473's discussion covers many more topics, if we go back to the title and first question in the issue I think this resolves #473.

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -953,7 +953,9 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 
 #### `NICK` (Nickname) `g7:NICK`
 
-A descriptive or familiar name that is used instead of, or in addition to, one’s proper name.
+A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
+A nickname can be a derivation of another name, such as "Jimmy" fron "James", or a distinct name.
+Nicknames can be brief-lived and informal or used consistently even in formal situations throughout one's life.
 
 #### `NMR` (Number of marriages) `g7:NMR`
 
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -963,8 +963,11 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
 
 :::note
-Although the label of this structure is "nickname", the definition is limited to descriptive or familiar names, not to other name types sometimes called "nicknames" in English, such as names used in formal settings but not on legal documents.
-However, some applications show users only the text "nickname", so user-entered `NICK` payloads that go beyond the structure's definition may be found in some GEDCOM files.
+The title and description of this structure was introduced with version 5.5 in 1996, but are understood differently by different users.
+Some use NICK only for names that would be inappropriate in formal settings.
+Some use it for pseudonyms regardless of where they are used.
+Some use it for any variant of a name that is not the one used on legal documents 
+Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the NICK structure is possible without contradicting some existing data.
 :::
 
 #### `NMR` (Number of marriages) `g7:NMR`
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -963,11 +963,11 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
 
 :::note
-The title and description of this structure was introduced with version 5.5 in 1996, but are understood differently by different users.
-Some use NICK only for names that would be inappropriate in formal settings.
+The label "nickname" and description text of this structure were introduced with version 5.5 in 1996, but are understood differently by different users.
+Some use `NICK` only for names that would be inappropriate in formal settings.
 Some use it for pseudonyms regardless of where they are used.
-Some use it for any variant of a name that is not the one used on legal documents 
-Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the NICK structure is possible without contradicting some existing data.
+Some use it for any variant of a name that is not the one used on legal documents.
+Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the `NICK` structure is possible without contradicting some existing data.
 :::
 
 #### `NMR` (Number of marriages) `g7:NMR`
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -953,7 +953,9 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 
 #### `NICK` (Nickname) `g7:NICK`
 
-A descriptive or familiar name that is used instead of, or in addition to, one’s proper name.
+A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
+A nickname can be a derivation of another name, such as "Jimmy" fron "James", or a distinct name.
```
tychonievich, line 957:
```suggestion
A nickname can be a derivation of another name, such as "Jimmy" from "James", or a distinct name.
```

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -953,7 +953,9 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 
 #### `NICK` (Nickname) `g7:NICK`
 
-A descriptive or familiar name that is used instead of, or in addition to, one’s proper name.
+A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
+A nickname can be a derivation of another name, such as "Jimmy" from "James", or a distinct name.
+Nicknames can be brief-lived and informal or used consistently even in formal situations throughout one's life.
```
tychonievich, line 958: The definition text in the spec doesn't use "Nickname", instead using "descriptive or familiar", so this added text may not be appropriate

tychonievich (commented):
Changes needed; the tag wasn't defined in terms of nickname before.

dthaler (dismissed):
See Luther's comment for changes needed.

specification/gedcom-3-structures-3-meaning.md
tychonievich, line 958:
```suggestion
A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.

:::note
Although the label of this structure is "nickname", the definition is limited to descriptive or familiar names, not to other name types sometimes called "nicknames" in English, such as names used in formal settings but not on legal documents.
However, some applications show users only the text "nickname", so user-entered `NICK` payloads that go beyond the structure's definition may be found in some GEDCOM files.
:::
```

dthaler (approved):

tychonievich:
Discussed in steering committee 25 JUN 2024

We identified former US president Jimmy Carter as having the nickname Jimmy, but Jimmy was arguably not "descriptive or familiar" (though arguably it was familiar, just familiar to everyone). Likewise Tippecanoe as a nickname for William Henry Harrison, another US president example of a publicly-used nickname that was not defined from any given name.

We generally agreed that we'd probably record these as a NICK, but that they aren't a great match for its defining text and thus that perhaps the definition of NICK should be expanded. But we did not settle on how it should be expanded, nor discuss if such an expansion is a patch.

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -960,7 +960,12 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
 
 #### `NICK` (Nickname) `g7:NICK`
 
-A descriptive or familiar name that is used instead of, or in addition to, one’s proper name.
+A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
+
+:::note
+Although the label of this structure is "nickname", the definition is limited to descriptive or familiar names, not to other name types sometimes called "nicknames" in English, such as names used in formal settings but not on legal documents.
+However, some applications show users only the text "nickname", so user-entered `NICK` payloads that go beyond the structure's definition may be found in some GEDCOM files.
+:::
```
tychonievich, line 968:
```suggestion
:::note
The title and description of this structure was introduced with version 5.5 in 1996, but are understood differently by different users.
Some use NICK only for names that would be inappropriate in formal settings.
Some use it for pseudonyms regardless of where they are used.
Some use it for any variant of a name that is not the one used on legal documents 
Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the NICK structure is possible without contradicting some existing data.
:::
```

tychonievich: My goal here is to have text that acknowledges the diversity of use without suggesting that one or another use is "proper"

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -960,7 +960,15 @@ See also `INDIVIDUAL_ATTRIBUTE_STRUCTURE`.
...
-A descriptive or familiar name that is used instead of, or in addition to, one’s proper name.
+A descriptive or familiar name that is used instead of, or in addition to, one’s official or legal name.
+
+:::note
+The title and description of this structure was introduced with version 5.5 in 1996, but are understood differently by different users.
+Some use NICK only for names that would be inappropriate in formal settings.
+Some use it for pseudonyms regardless of where they are used.
+Some use it for any variant of a name that is not the one used on legal documents 
+Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the NICK structure is possible without contradicting some existing data.
+:::
```
tychonievich, line 971:
```suggestion
:::note
The label "nickname" and description text of this structure were introduced with version 5.5 in 1996, but are understood differently by different users.
Some use `NICK` only for names that would be inappropriate in formal settings.
Some use it for pseudonyms regardless of where they are used.
Some use it for any variant of a name that is not the one used on legal documents.
Because all of these uses, and likely others as well, are common in existing data, no further clarification of the meaning of the `NICK` structure is possible without contradicting some existing data.
:::
```

PR #492, opened 2024-06-25
Define "principle date"

### LINKED ISSUE #490: Source_Citation SOUR.DATA.DATE has no definition.
In GEDCOM v5.5.1 the definition for `SOUR.DATA.DATE` was: <ENTRY_RECORDING_DATE>
>The date that this event data was entered into the original source document.

This definition at a minimum is missing from the the current version of GEDCOM v7.0.14

For Source_Citations this date definition plus the date you last visited the website is also be needed.

A new definition for the payload could include both: 
>"For immutable source documents, (i.e. Census, Historical Church Documents) the date represents when this data was entered in the original source document, for variable and subject to change sources (i.e. websites, aggregator documents) the date represents when the data was last visited."

tychonievich:
As pointed out in #488 and #490, the definition of DATE is unclear. In particular, it hinges on the the vague phrase "principle date" which could use some clarification. This is my effort to provide that clarification.

Note, if competing definitions of the principle date in these contexts exists then this suggestion could be seen as backwards-incompatible and may need to be reworded as a non-normative recommendation or note. That said, I'm not aware of any conflicting definitions.

Resolves #490

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -437,6 +437,13 @@ with substructures providing additional details about the source (not the export
 The principal date of the subject of the superstructure.
 The payload is a `DateValue`.
 
+When the superstructure is an event, the principle date indicates when the event took place.
+
+When the superstrucutre is an attribute, the principle date indicates when the attribute was observed, asserted, or applied.
+A date period might put bounds on the attributes applicability, but other date forms assume that the attribute may have also applied on other dates too.
+
+When the superstructure is a `g7:SOUR-DATA`, the principle date indiates when the data was entered into the source; or, for a source like a website that changes over time, a date on which the source contained the data.
+
 See `DATE_VALUE` for more details.
 
 #### `DATE` (Date) `g7:DATE-exact`
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -437,12 +437,12 @@ with substructures providing additional details about the source (not the export
 The principal date of the subject of the superstructure.
 The payload is a `DateValue`.
 
-When the superstructure is an event, the principle date indicates when the event took place.
+When the superstructure is an event, the principal date indicates when the event took place.
 
-When the superstrucutre is an attribute, the principle date indicates when the attribute was observed, asserted, or applied.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -437,6 +437,13 @@ with substructures providing additional details about the source (not the export
 The principal date of the subject of the superstructure.
 The payload is a `DateValue`.
 
+When the superstructure is an event, the principle date indicates when the event took place.
+
+When the superstrucutre is an attribute, the principle date indicates when the attribute was observed, asserted, or applied.
+A date period might put bounds on the attributes applicability, but other date forms assume that the attribute may have also applied on other dates too.
+
+When the superstructure is a `g7:SOUR-DATA`, the principle date indiates when the data was entered into the source; or, for a source like a website that changes over time, a date on which the source contained the data.
+
```
tychonievich, line 446:
```suggestion
When the superstructure is an event, the principal date indicates when the event took place.

When the superstructure is an attribute, the principal date indicates when the attribute was observed, asserted, or applied.
A date period might put bounds on the attributes applicability, but other date forms assume that the attribute may have also applied on other dates too.

When the superstructure is a `g7:SOUR-DATA`, the principle date indicates when the data was entered into the source; or, for a source like a website that changes over time, a date on which the source contained the data.

```

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -437,6 +437,13 @@ with substructures providing additional details about the source (not the export
 The principal date of the subject of the superstructure.
 The payload is a `DateValue`.
 
+When the superstructure is an event, the principal date indicates when the event took place.
+
+When the superstructure is an attribute, the principal date indicates when the attribute was observed, asserted, or applied.
+A date period might put bounds on the attributes applicability, but other date forms assume that the attribute may have also applied on other dates too.
+
+When the superstructure is a `g7:SOUR-DATA`, the principle date indicates when the data was entered into the source; or, for a source like a website that changes over time, a date on which the source contained the data.
+
```
tychonievich, line 446:
```suggestion
When the superstructure is a `g7:SOUR-DATA`, the principal date indicates when the data was entered into the source; or, for a source like a website that changes over time, a date on which the source contained the data.

```

PR #506, opened 2024-07-09
Restore 3.0's definition of jurisdiction

### LINKED ISSUE #496: Definition of PLAC
It should be clearly defined what the lowest unit of jurisdiction is. Could a city district, a street, a building, a house be part of PLAC or is it part of ADDR? What about a hospital, a cemetery, a church?

tychonievich:
Restored the definition of "jurisdiction" that was present in version 3.0.

Resolves #496

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1121,7 +1121,8 @@ A name given to a foundling orphan might be
 
 The principal place in which the superstructure's subject occurred,
 represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction,
-where each jurisdiction is a unit in a political, ecclesiastical, or geographical hierarchy.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -1120,7 +1120,8 @@ A name given to a foundling orphan might be
 #### `PLAC` (Place) `g7:PLAC`
 
 The principal place in which the superstructure's subject occurred,
-represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction.
+represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction,
+where each jurisdiction is a unit in a political, ecclesiastical, or geographical hierarchy.
```
tychonievich, line 1124:
```suggestion
where "jurisdiction" includes units in a political, ecclesiastical, and geographical hierarchies
and may include units as large as a nation or as specific as a building, farm, or cemetery.
```

PR #514, opened 2024-07-16
Increase largest PLAC example

### LINKED ISSUE #512: PLAC jurisdiction can be larger than a nation
Text currently reads:
> The principal place in which the superstructure's subject occurred, represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction, where "jurisdiction" includes units in a political, ecclesiastical, and geographical hierarchies and may include units **as large as a nation** or as specific as a building, farm, or cemetery. As with other lists, the jurisdictions are separated by commas. Any jurisdiction's name that is missing is still accounted for by an empty string in the list.

The "as large as a nation" can be read as saying that's the largest possible jurisdiction.
However there are records that just list continent (e.g., "emigrated to North America"), or ocean.

tychonievich:
resolves #512

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1122,7 +1122,7 @@ A name given to a foundling orphan might be
 The principal place in which the superstructure's subject occurred,
 represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction,
 where "jurisdiction" includes units in a political, ecclesiastical, and geographical hierarchies
-and may include units as large as a continent or "at sea", or as specific as a building, farm, or cemetery.
+and may include units of any size, such as a continent, "at sea", or a specific building, farm, or cemetery.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -1122,7 +1122,7 @@ A name given to a foundling orphan might be
 The principal place in which the superstructure's subject occurred,
 represented as a [List] of jurisdictional entities in a sequence from the lowest to the highest jurisdiction,
 where "jurisdiction" includes units in a political, ecclesiastical, and geographical hierarchies
-and may include units as large as a nation or as specific as a building, farm, or cemetery.
+and may include units as large as a continent or "at sea", or as specific as a building, farm, or cemetery.
```
tychonievich, line 1125:
```suggestion
and may include units of any size, such as a continent, "at sea", or a specific building, farm, or cemetery.
```

PR #527, opened 2024-08-06
Adding SPLAC under PLAC

dthaler:
Conversation draft of adding place records to 7.1

This puts SPLAC as a substructure of PLAC, like PERSONAL_NAME_PIECES are under PERSONAL_NAME_STRUCTURE.  See PR #520 for another alternative, and diffs between the two proposals can be seen [here](https://github.com/FamilySearch/GEDCOM/compare/splac-beside-plac...splac-under-plac).

Putting SPLAC under (rather than alongside/instead of) PLAC is proposed in this PR for two reasons:

1. For consistency with actual usage of [GEDCOM-L's _LOC extension](https://genealogy.net/GEDCOM/GEDCOM551%20GEDCOM-L%20Addendum-R2.pdf#page=13).
2. So that a FamilySearch GEDCOM 7.0 (or even 5.5.1) application can read a 7.1 GEDCOM file using SPLAC without losing the place names.   The analogy is similar to how name parts can be optionally used whereas applications that don't support them still use the NAME payload.

This PR only addresses the record-vs-substructure topic. The additional substructures that have also been considered for location records (for example, see [GEDCOM-L's _LOC extension](https://genealogy.net/GEDCOM/GEDCOM551%20GEDCOM-L%20Addendum-R2.pdf#page=13)) would presumably be added to the new <<PLACE_DETAILS>> production in a future PR if we decide that this organization is the right approach.

### DIFF
--- a/specification/gedcom-3-structures-1-organization.md
+++ b/specification/gedcom-3-structures-1-organization.md
@@ -415,16 +415,15 @@ If a city is part of a county which is part of a state, the city's place record 
 Multiple `<<SHARED_PLACE_STRUCTURE>>`s are permitted to support places within multiple hierarchies (for example, a church that's both within an ecclesiastical region and a political region).
 
 Shared place records offer more flexibility than place structures do.
-A `PLAC` can be replaced by an `SPLAC` without loss of information by making one `SPLAC` record for each non-empty string in the `PLAC`'s payload and linking them together using the `SPLAC`'s `SHARED_PLACE_STRUCTURE`s.
+A `PLAC` can be augmented by an `SPLAC` by making one `SPLAC` record for each non-empty string in the `PLAC`'s payload and linking them together using the `SPLAC`'s `SHARED_PLACE_STRUCTURE`s.
 Information is copied into the new chain of `SPLAC` records as follows:
 
-- The `PLAC` structure is replaced by an `SPLAC` structure that points to the first `SPLAC` record in the chain
-- Any `NOTE` substructures of `PLAC` are retained as substructures of the new `SPLAC` structure
+- An `SPLAC` substructure is added to the `PLAC` that points to the first `SPLAC` record in the chain.
 - Empty list entries are skipped.
-- The `PLAC` payload parts become `SPLAC` payloads
-- The `FORM` payload parts (which may be copied from the `HEAD`.`PLAC`.`FORM` if that is present but `PLAC`.`FORM` is not) become `SPLAC`.`TYPE` payloads
-- The `TRAN` payload parts become `TRAN` payloads
-- `LANG` and `TRAN`.`LANG` are copied to each record in the linked list
+- The `PLAC` payload parts become `SPLAC` payloads.
+- The `FORM` payload parts (which may be copied from the `HEAD`.`PLAC`.`FORM` if that is present but `PLAC`.`FORM` is not) become `SPLAC`.`TYPE` payloads.
+- The `TRAN` payload parts become `TRAN` payloads.
+- `LANG` and `TRAN`.`LANG` are copied to each record in the linked list.
 - All other substructures (`MAP` and `EXID`) are copied only to the first record in the list.
 
 :::example
@@ -443,10 +442,11 @@ The 7.0 structure
 3 NOTE this is an example
 ```
 
-can be converted without loss of information into the new structure
+can be augmented, becoming the new structure
 
 ```gedcom
-2 SPLAC @SP1@
+2 PLAC one, two, , three
+3 SPLAC @SP1@
 3 NOTE this is an example
 ```
 
@@ -477,9 +477,6 @@ and new records
 
 :::
 
-Conversion in the other direction is also possible (from `SPLAC` to `PLAC`) but will in general discard information about individual places in the `SPLAC` chain.
-
-
 #### `SOURCE_RECORD` :=
 
 ```gedstruct
@@ -488,7 +485,7 @@ n @XREF:SOUR@ SOUR                         {1:1}  g7:record-SOUR
      +2 EVEN <List:Enum>                   {0:M}  g7:DATA-EVEN
         +3 DATE <DatePeriod>               {0:1}  g7:DATA-EVEN-DATE
            +4 PHRASE <Text>                {0:1}  g7:PHRASE
-        +3 <<PLACE_REFERENCE>>             {0:1}
+        +3 <<PLACE_STRUCTURE>>             {0:1}
      +2 AGNC <Text>                        {0:1}  g7:AGNC
      +2 <<NOTE_STRUCTURE>>                 {0:M}
   +1 AUTH <Text>                           {0:1}  g7:AUTH
@@ -651,7 +648,7 @@ Time phrases are expected to be added in version 7.1.
 
 ```gedstruct
 n <<DATE_VALUE>>                           {0:1}
-n <<PLACE_REFERENCE>>                      {0:1}
+n <<PLACE_STRUCTURE>>                      {0:1}
 n <<ADDRESS_STRUCTURE>>                    {0:1}
 n PHON <Special>                           {0:M}  g7:PHON
 n EMAIL <Special>                          {0:M}  g7:EMAIL
@@ -1041,7 +1038,7 @@ Ordinances performed by members of The Church of Jesus Christ of Latter-day Sain
 ```gedstruct
 n <<DATE_VALUE>>                         {0:1}
 n TEMP <Text>                            {0:1}  g7:TEMP
-n <<PLACE_REFERENCE>>                    {0:1}
+n <<PLACE_STRUCTURE>>                    {0:1}
 n STAT <Enum>                            {0:1}  g7:ord-STAT
   +1 DATE <DateExact>                    {1:1}  g7:DATE-exact
      +2 TIME <Time>                      {0:1}  g7:TIME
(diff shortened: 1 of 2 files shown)


tychonievich:
Discussed in steering committee 6 AUG 2024

- Those present were happy with both this and #520, but would want input from the GEDCOM-L community that uses the SPLAC-like `_LOC` extension now before committing to either one.
- This proposal has the benefit of being more like how `_LOC` works (a substructure of `PLAC`).
- The other proposal has the benefit of a clearer path towards deprecating `PLAC`.
- Either #520 and #527 would also want to have more substructures added before being released.

tychonievich:
The GEDCOM-L discussed this and reported back to the steering committee. If SPLAC is added in 7.1, they prefer this solution to #520. But they'd rather minimize the number of times things change, so if we plan to have something unlike this in 8.0 (i.e. remove `PLAC` entirely), then they'd prefer to not do anything in 7.1 and wait until 8.0.

dthaler:
Discussion in GEDCOM Steering Committee 13 AUG 2024:
* GEDCOM-L discussed issues with merging between one file using only SPLAC and another using only PLAC, in the other proposal which has no specific association between PLAC and SPLAC values, whereas this proposal doesn't have that issue since SPLAC requires PLAC still exist.
* Legacy and TNG and MagiCensus are other apps that store per-place (not per instance of usage) data in GEDCOM files so additional input would be useful.  (MagiCensus developers have no strong opinion here and are ok with either option discussed by the committee.)
* It would be good to see other PRs that add the other capabilities from the `_LOC` extension (date, etc.) to complete the proposals.
* It would be good to see a v8 proposal draft that removes `PLAC`, to allow getting Legacy, TNG, etc. to weigh in across the set.

Norwegian-Sardines:
Luther,  Can you please explain the discussion better?
First, I was hoping that I would be party to the discussion on PLAC vs SPLAC, I was asked to attend but have been camping for the last 6 weeks and been away from regular internet access during that time.

I had hoped that SPLAC would stand along side PLAC in v7.1 so that either open could be used, and later the PLAC tag could be deprecated.

One of the reasons I've disliked SPLAC (and _LOC) as a substructure to PLAC is when the PLAC value does not match the dates parameters of the _LOC structure.  Which is correct?  What happens when the _LOC hierarchy is corrected when a a location changes its parent but the PLAC does not change.  Have the data in two places, possibly used for different reasons (1) Historical accuracy, (2) Modern Map Linking to Mapping software?

dthaler:
Discussion in GEDCOM Steering Committee 20 AUG 2024:
* The committee discussed examples like @Norwegian-Sardines mentioned above where PLAC and SPLAC could get out of sync in terms of the hierarchy. 
* If you just put the lowest level in the PLAC payload then this problem doesn't occur, but it is not backwards compatible without data loss.
* If information is in both places for backwards compatibility, there is always the danger of being out of sync with each other.  One could perhaps declare a precedence between the two as to which one wins, but it's unclear if that's helpful.
* If information is not in both places, one loses backwards compatibility with apps/websites that don't yet support the new structure.
* We want this to be a topic of discussion in a public meeting that will be scheduled before making any decisions. Watch the discussion board for details. The purpose of these draft PRs is to generate such discussion.

fisharebest:
A supposed "advantage" of the hierarchical `_LOC` structure is the ability to add dates to the relationships.  Thus you can represent situtations such as Gdansk being variously in Poland, Germany, Russia, Prussia, etc.

I'm going to argue that this is a disadvantage, not an advantage.

By allowing this "historical geographic database" to be included in a GEDCOM file, you are effectively *requiring* it.  If I want to record events in Gdansk, I need to know/include the exact dates for all of its history.

Now every GEDCOM file will needs to contain the same historic geographic database.

This feels completely wrong.  The historic geographic database ought to be part of a genealogy application - not included in every genealogy file.

IMHO, `PLAC` (and `SPLAC`, etc.) should contain the current/contemporary place names.  Historic place names would be used in source-citations etc.

Norwegian-Sardines:
> A supposed "advantage" of the hierarchical `_LOC` structure is the ability to add dates to the relationships. Thus you can represent situtations such as Gdansk being variously in Poland, Germany, Russia, Prussia, etc.
> 
> I'm going to argue that this is a disadvantage, not an advantage.
> 
> By allowing this "historical geographic database" to be included in a GEDCOM file, you are effectively _requiring_ it. If I want to record events in Gdansk, I need to know/include the exact dates for all of its history.
> 
> Now every GEDCOM file will needs to contain the same historic geographic database.
> 
> This feels completely wrong. The historic geographic database ought to be part of a genealogy application - not included in every genealogy file.
> 
> IMHO, `PLAC` (and `SPLAC`, etc.) should contain the current/contemporary place names. Historic place names would be used in source-citations etc.

I understand your concern, and it is a valid to a point.  We are trained to record the exact "place" an individual was born/died.  So if a person was born in Gdansk, Prussia and an ancestor was born in Gdansk, Poland you are saying that both individuals would have Gdansk, Poland as their birth place, giving a wrong impression of the fact for a reader 10 years down the road who only has access to the report or book produced out of the GEDCOM.

So maybe a better concept is not not have a hierarchical concept at all, but a concept where the `PLAC` tag holds the recorded place and an optional link to a "Historical_Place_Record" (`HPLAC`) that is a single record which contains detail about the place listing all of its names, dates and history.  This would allow the `PLAC` to record the proper birth place used in archived reports/books, and if present the `HPLAC` would tell readers the coordinates, alternate name list (with dates?) and a short history with pictures of the place for people who want to know about the place.  

Cities, Towns, Countries, and other places can have varied names over time (the town my grandfather was born has changed names 3 times since he was born there and move location once) and that information needs to be recorded (the birth town) and explained for readers down the road, particularly when the birth certificate says one thing and the book says another!

dthaler:
> A supposed "advantage" of the hierarchical `_LOC` structure is the ability to add dates to the relationships. Thus you can represent situtations such as Gdansk being variously in Poland, Germany, Russia, Prussia, etc.
> 
> I'm going to argue that this is a disadvantage, not an advantage.
> 
> By allowing this "historical geographic database" to be included in a GEDCOM file, you are effectively _requiring_ it. If I want to record events in Gdansk, I need to know/include the exact dates for all of its history.
> 
> Now every GEDCOM file will needs to contain the same historic geographic database.
> 
> This feels completely wrong. The historic geographic database ought to be part of a genealogy application - not included in every genealogy file.

Interesting... I understand the above argument (which seems valid to me) to be arguing against an `SPLAC` record and instead just using the mechanism we already have in 7.0 of using `PLAC`.`EXID` with (say) a FamilySearch Place ID as registered at https://gedcom.io/_pages/exid-type/ as the "historic geographic database".  That is, the database itself need not be "part of" the genealogy application per se, as long as the application can access the database online, which the existing `EXID` does enable.  Of course there could be other historic geographic databases besides the FamilySearch Place one, and one would then want an `EXID` type registered for whatever else too.  But I understand you're arguing to do that and not add that database into each GEDCOM file itself.  That seems like a valid point to me, as an implementer.  (FWIW, my app stores the familysearch ID of the place in GEDCOM 5.5.1 files using an extension, so your suggestion in a sense meets the "Used" criteria just as the `_LOC` extension used by other apps also does.)

> IMHO, `PLAC` (and `SPLAC`, etc.) should contain the current/contemporary place names. Historic place names would be used in source-citations etc.

I think this point is orthogonal to the rest of the point.  And this one I disagree with... some users would enter the place as it was in the record (i.e., the historic name) and other users would enter the place using the current place name.  Since the GEDCOM 7.0 (and earlier) specs did not say either way, that means I would argue we must permit both uses until 8.0 (the earliest we could make a breaking change) and hence you can never tell from the PLAC payload which name it is - current vs historic.   Once you use the EXID to point to a historic geographic database entry though, you can look up the name as of any given date and solve the problem though.

Norwegian-Sardines:
>I think this point is orthogonal to the rest of the point. And this one I disagree with... some users would enter the place as it was in the record (i.e., the historic name) and other users would enter the place using the current place name. Since the GEDCOM 7.0 (and earlier) specs did not say either way, that means I would argue we must permit both uses until 8.0 (the earliest we could make a breaking change) and hence you can never tell from the PLAC payload which name it is - current vs historic. Once you use the EXID to point to a historic geographic database entry though, you can look up the name as of any given date and solve the problem though.

See my comment above about `HPLAC` linked record!

albertemmerich:
The possibility to enter historical names and administrative levels does *not* imply, that you must do that. There are a lot of other structures in GEDCOM which are only found in specialised applications. You can do it as you did so far - trying to get todays name (and administrations levels) of a place. 

However an example: You have a village of maybe 20 farms. You will not find a historical gazetteer showing those details. However the place records we already implemented by GEDCOM-L location records give you the possibility to document the names of the farms, their exact locations, their type and so on - even varying over time.

@fisharebest: You will write Калининград in Russia, if the source tells you Königsberg in Preußen? And sources do not say: he was born in Gdansk, Prussia. The city was called Danzig, which still is its German name. So it was Danzig in Preußen... Place records can manage the names of places varying in time and by language. Munich in Bavaria is the same city as München in Bayern, to have another example only based on different languages.

See:

- http://gov.genealogy.net/item/show/KONSSEKO04GR for Königsberg in Preußen,
- https://gov.genealogy.net/item/show/DANZIGJO94HI for Danzig,
- https://gov.genealogy.net/item/show/MUNHENJN58SD for München

Norwegian-Sardines:
The `PLAC`.`EXID` to sites like "FamilySearch" is not good enough as a "place history" when we include places that are not in any external database.  Places not limited to farms, estates, hospitals, small cemeteries, but could also be places with an expanded history that is important to a researcher and their readers concerning a small rural community, or even larger cities that have expanded to include neighboring communities.  Sharing, merging data from multiple researcher may want to have these expanded histories of little know places included in the larger context.  Many of the most important places to my family history are not found in any external database or the data is very limited!

dthaler:
Based on https://github.com/FamilySearch/GEDCOM/discussions/652 this is for 8.0 but preference may be to remove PLAC all together in 8.0.

dthaler:
I prefer removing PLAC in 8.0 so abandoning this proposal for now.

Norwegian-Sardines:
Albert said:
@fisharebest: You will write Калининград in Russia, if the source tells you Königsberg in Preußen? And sources do not say: he was born in Gdansk, Prussia. The city was called Danzig, which still is its German name. 

Having just been in Gdansk, I know that many Polish people hated the name Danzig and very often wrote/said Gdansk even when it was called Danzig, Prussia by the Germans!  When capturing data for individuals not in my immediate family I may write the data as presented to me rather than getting the "Sources" that verify a situation or where no "source" exists for a particular "fact" other than hearsay!

But rather than picking on this specific example understand that place names change, places can have two names at the same time depending on the ethnicity or religion of the person making the statement or creating the source.  Not all place names are exact when captured!

PR #557, opened 2024-10-12
Update generate-files to use a PAT

### LINKED ISSUE #503: generate-files github workflow files pull requests that are blocked because validate-yaml never runs
See https://github.com/orgs/community/discussions/57484 for explanation.

https://github.com/FamilySearch/GEDCOM/blob/main/.github/workflows/generate-files.yml#L72 current uses the GITHUB_TOKEN which is the problem.

The recommended fix seems to be to use a PAT rather than GITHUB_TOKEN.

dthaler:
* Fix names of workflows
* Make generate-files use a PAT so that validate-yaml will run

The PAT is already generated and configured in the repository secrets

Fixes #503


dthaler:
Approved during GEDCOM steering committee meeting

PR #559, opened 2024-10-13
Don't quote NO (or YES just for consistency)

### LINKED ISSUE #558: Wrong use of quote marks in one of the YAML files?
The standard tag In the YAML file at https://gedcom.io/terms/v7/NO is quoted with single quote marks ie. it is 'NO' instead of NO. I think this might be a mistake and it is causing me a problem with some code I am writing. Other standar tags are not quoted.

dthaler:
Fixes #558

### DIFF
--- a/build/uri-def.py
+++ b/build/uri-def.py
@@ -396,7 +396,7 @@ if __name__ == '__main__':
             
             if g7[tag][0] in ('structure', 'enumeration', 'calendar', 'month'):
                 ptag = re.sub(r'.*-', '', re.sub(r'-[A-Z]?[a-z].*', '', tag))
-                print('\nstandard tag: '+(repr(ptag) if ptag in ('YES','NO','TRUE','FALSE', '0', '1', '2', '3') else ptag), file=fh)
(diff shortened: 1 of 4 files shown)


build/uri-def.py
```diff
@@ -396,7 +396,7 @@ def expand_prefix(txt, prefixes):
             
             if g7[tag][0] in ('structure', 'enumeration', 'calendar', 'month'):
                 ptag = re.sub(r'.*-', '', re.sub(r'-[A-Z]?[a-z].*', '', tag))
-                print('\nstandard tag: '+(repr(ptag) if ptag in ('YES','NO','TRUE','FALSE', '0', '1', '2', '3') else ptag), file=fh)
+                print('\nstandard tag: '+(repr(ptag) if ptag in ('TRUE','FALSE', '0', '1', '2', '3') else ptag), file=fh)
```
tychonievich, line 399: We cannot remove quotes from strings that are also ways of spelling Boolean values without changing semantics. This suggested change breaks YAML.

If we want to be consistent in where quotes appear, we could instead make an unconditional use of `repr`

tychonievich (changes requested):

PR #560, opened 2024-10-15
Always quote tags

tychonievich:
Resolves #559


dthaler:

PR #562, opened 2024-10-22
Don't warn about use of "on" as a YAML key

dthaler:
Per documentation at
https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy


dthaler:

PR #580, opened 2025-01-01
Personal names are not required to start with a "1".

sharth:
This is a minor error in the specification. By using `1*nameChar`, this would mean that "John Doe" is illegal, but "1John Doe" is okay.

This is unlikely to be intentional.


tychonievich:
The `1*` is intentional; as noted in the summary of ABNF in section 1.1,
> Preceding a group or element by `*` means any number may be included.
> Preceding a group or element by `1*` means 1 or more may be included.

Thus, `1*nameChar` means a string of one more more `nameChar`s: that is, empty strings are not valid names, but any non-empty string (not containing slashes or tabs) are valid. Removing the 1 would allow empty strings as names.

sharth:
Whoops. I think I thought this was more like EBNF. Thanks for the time.

PR #590, opened 2025-02-04
Update how tables with URIs in them are parsed

### LINKED ISSUE #574: build/uri-def.py failing to extract some definitions
In answering #573 I realized that g7:INDI-TITL is defined [in the spec](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#individual-attributes) as

> A formal designation used by an individual in connection with positions of royalty or other social status, such as Grand Duke.

but that text is not successfully extracted by uri-def.py into [the corresponding YAML file](https://gedcom.io/terms/v7/INDI-TITL).

tychonievich:
Resolves #574


tychonievich:
There's still an error; the spec refers to `g7:enum-RELI` but the file we have in the `g7:` directory is called `RELI` instead. This is not a full fix. I'll try to find a fix for the other part soon.

dthaler (approved):

tychonievich:
enum- placement error also fixed (including in a few places I'd not noticed it before)

dthaler (approved):

Issue #593, opened 2025-02-14
Encouragement for using more HTML not explicitly noted in g7:MIME


Norwegian-Sardines:
The v7.0 (and probably v7.1) of the GEDCOM Standard says the following in the g7:MIME definition.

> Supporting more of HTML is encouraged. Unsupported elements should be ignored during display.

The statement, "Supporting more of HTML is encouraged." may need some written restrictions if kept in future releases of GEDCOM.

For online applications that encounter HTML in the transmission payload any "unsupported elements" should probably be removed from the display (rather than just ignored) and will most likely be lost if re-transmitted back to any other desktop application.  The concept of encouraging the use of HTML that is:

1) Not strictly for display (code injection, scripts, input) should be discouraged
2) Changes the online display design such as font size/color as well as advanced HTML (Audio, Canvas, Page Navigation) can be ignored, but may also be lost during use in an online application.

The strict understanding of the use of "should be ignored during display" may be misunderstood to mean (still displayed) in a desktop based application, but may be removed from the payload in an online application.

Should the above quote be reviewed with these concerns in mind?  Should it be dropped all together?

dthaler:
Discussion during GEDCOM Steering Committee 18 FEB 2025:
* We should probably change "Unsupported elements should be ignored during display." to "Unsupported tags should be ignored during display." since later in the same section it says "Remove all other `<…>` tags" and the intent is the resulting display should be equivalent to if the unsupported tags were removed.
* We did discuss the possibility of dropping the sentence all together but it may be clearer to be explicit.
* We have not found any existing specification for what tags are "strictly for display" and would prefer not to define it ourselves.
* The `<table>` structure is an example that some apps might support and others might not and those that don't will display the result oddly.

PR #594, opened 2025-02-18
replace "entities" with "tags" in HTML

tychonievich:
Addresses an issue raised in #593


dthaler (approved):

dthaler:

PR #616, opened 2025-03-25
Update language for AGE

### LINKED ISSUE #610: Confusing "Age in the document" text
https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#AGE defines AGE as

> The age of the individual at the time an event occurred, or the age listed in the document.

However, AGE cannot appear beneath a SOUR and has no other way to indicate which document is meant by "the age listed in the document." This text did not appear in 5.5.1. I recommend we remove ", or the age listed in the document" entirely.

dthaler:
Fixes #610

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -226,7 +226,7 @@ See `ADDRESS_STRUCTURE` for more details.
 
 #### `AGE` (Age at event) `g7:AGE`
 
-The age of the individual at the time an event occurred, or the age listed in the document.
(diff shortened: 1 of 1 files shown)


Norwegian-Sardines:
Not knowing which Source Artifact (“SA”) produced the AGE payload is not any different than when asserting a DATE or PLAC payload which are also based on a SA! A “fact” could have multiple SAs with one asserting a place, another asserting a date and a third asserting an age! Thus, if AGE requires an associated SA the DATE and PLAC would as well!

tychonievich (approved):

dthaler:

PR #625, opened 2025-04-08
Allow other media types as extensions

### LINKED ISSUE #622: Is adding new NOTE.MIMEs permitted by 7.0?
The definition of [`g7:MIME`](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#MIME) says:

> As of version 7.0, only 2 media types are supported by this structure:

and then lists text/plain and text/html. In a recent conversation with @dthaler we discovered we interpreted "supported" differently:

I think it is broadly a synonym for "recommended" or "documented" and that using a different media type (like text/markdown) in a 7.0 file is OK, though perhaps not likely to be widely supported by other tools (being something like an extension).

I understand @dthaler to think it is broadly a synonym for "allowed" and that if a different media type is included then the file is no longer valid 7.0.

We should agree on what is meant and chance the language to be less ambiguous.

dthaler:
Fixes #622

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -905,7 +905,7 @@ for this asset, the `FORM`.`MEDI` is recommended to be `PHOTO` rather than `ELEC
 
 Indicates the [media type](#media-type) of the payload of the superstructure.
 
-As of version 7.0, only 2 media types are supported by this structure:
+As of version 7.0, there are two standard media types for this structure:
 
 - `text/plain` shall be presented to the user as-is, preserving all spacing, line breaks, and so forth.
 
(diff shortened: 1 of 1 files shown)


chronoplexsoftware:
The proposals seem to permit (though not recommend) any media type to be used. Surely that significantly alters the concept of a note which is intended to be textual representations of information.

    2.1. Text
    A free-text string is text in a human language. Conceptually, it may be either a user-generated string or
    a source-generated string. Programmatically, both are treated as unconstrained sequences of characters
    with an associated language.

At the very least, should this not be constrained to MIMEs of type text?

Subtypes other than plain and all those suggested to date seem to have all been requested on the basis of supporting formatting and this should probably be the principle reason for anyone to consider use of another subtype.

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -943,7 +943,12 @@ If needed, `text/html` can be converted to `text/plain` using the following step
 2. Case-insensitively replace each `<p`...`>`, `</p`...`>`, and `<br`...`>` with a line break
 3. Remove all other `<`...`>` tags
 4. Replace each `&lt;` with `<` and `&gt;` with `>`
-4. Replace each `&amp;` with `&`
+5. Replace each `&amp;` with `&`
+
+Other media types not discussed above are also permitted, though not recommended.
```
dthaler, line 948:
```suggestion
Other `text` media types not discussed above are also permitted, though not recommended.
```

dthaler:
Approved during GEDCOM Steering Committee meeting.

tychonievich (approved):

dthaler:
> At the very least, should this not be constrained to MIMEs of type text?

Yes, done.

PR #633, opened 2025-04-19
Update UUID reference

### LINKED ISSUE #629: Replace reference to RFC 4122 with RFC 9562
[RFC 4122](https://www.rfc-editor.org/rfc/rfc4122.txt), which is referenced in the definition of the `UID` structure, has been obsoleted by [RFC 9562](https://www.rfc-editor.org/rfc/rfc9562.txt).  The replacement is better written and fixes a number of technical errata in the original.  I can’t see any good reason not to update the reference.

If might also be worth tweaking the wording from ‘formatted as described in [RFC 4122](https://www.rfc-editor.org/info/rfc4122)’ to ‘formatted as described by the `UUID` production in [RFC 9562](https://www.rfc-editor.org/info/rfc9562)’.  Both RFCs define several serialisation formats, including as a URN (e.g. `urn:uuid:be8e67f3-4fcf-4fb2-bec6-0a3f2268a439`) as well as the bare UUID.  I assume GEDCOM 7.0 intends the bare UUID to be used, but it doesn’t say so explicitly.

dthaler:
Fixes #629

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1615,7 +1615,8 @@ Additional type URIs can be registered by filing a
 A globally-unique identifier of the superstructure,
 to be preserved across edits.
 If a globally-unique identifier for the record already exists, it should be used without modification, not even whitespace or letter case normalization.
-New globally unique identifiers should be created and formatted as described in [RFC 4122](https://www.rfc-editor.org/info/rfc4122).
+New globally unique identifiers should be created and formatted using the UUID
+production specified in [RFC 9562](https://www.rfc-editor.org/info/rfc9562) Section 4.
 
 This is metadata about the structure itself, not data about its subject.
 Multiple structures describing different aspects of the same subject would have different `UID` values.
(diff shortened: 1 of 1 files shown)


tychonievich (approved):

dthaler:

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -1615,7 +1615,8 @@ Additional type URIs can be registered by filing a
 A globally-unique identifier of the superstructure,
 to be preserved across edits.
 If a globally-unique identifier for the record already exists, it should be used without modification, not even whitespace or letter case normalization.
-New globally unique identifiers should be created and formatted as described in [RFC 4122](https://www.rfc-editor.org/info/rfc4122).
+New globally unique identifiers should be created and formatted using the UUID
```
dthaler, line 1618:
```suggestion
It is recommended that new globally unique identifiers be created and formatted using the UUID
```

dthaler:
Discussion during meeting:
Also clarified that this is just a recommendation since there are other UID payload formats already in the wild that do not follow this format, and the text already said existing ones should be left unmodified.  For consistency, the same rule also applies to new UID payload formats since applications already generate new ones with another format too.

PR #634, opened 2025-04-22
Clarify ELECTRONIC is used for a scan

### LINKED ISSUE #623: Add SCAN as Mediatype?
When entering sources and REPO's we can fill in the mediatype of the source.
Today a lot of media are scanned pages from books. They are called SCAN, even in Dutch texts.

Would it be a good idea to add SCAN to the list of media types?
I dont see it as a digital artefact, although it is digital ofcourse.

It is very common to say: I make a scan, and we put something in a scanner/printer to do so.

In earlier days things were put on microfiche, no idea if thats still done.
But the word scan is a very common word today.

dthaler:
Add descriptive phrase that specifically uses the word scan.
Fixes #623


dthaler:

PR #635, opened 2025-04-22
Explain use of UTF-8 file names in GEDZIP

### LINKED ISSUE #631: Must file names inside GEDZIP be UTF-8?
Section 2.12 says about the FILE payload:

> Syntactically, the payload is a URI reference as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), or a valid URL string as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).

RFC 3986 requires percent-encoding non-ascii characters, but do not require (even if it is the most common case) that http URIs use UTF-8 before percent encoding so could in theory be another encoding unless the GEDCOM spec says otherwise. Similarly the WHATWG spec says "Using anything but [UTF-8 decode without BOM](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom) when input contains bytes that are not [ASCII bytes](https://infra.spec.whatwg.org/#ascii-byte) might be insecure and is not recommended." So again UTF-8 is recommended but not required.

And the GEDZIP format uses the ZIP format (ISO/IEC 21320-1) which does not require UTF-8. The ISO spec explains:

> The “Zip” format was originally developed as a general purpose archive format, it had to support various platforms with various character encoding before ISO/IEC 10646 standardization. Therefore, Appnote
specifies few restrictions for filenames in the archive. For compatibility, this part of ISO/IEC 21320 does
not require additional restrictions on filenames which are valid according to Appnote.
However, some digital document packages using UTF-8 filenames in Zip archive do define restrictions
on filenames. In this Annex, these restrictions are summarized to help implementers avoid unexpected
information loss during the document transcoding.

General purpose bit flag 11 indicates whether it is restricted to UTF-8:

> If bit 11 is set, the file names and comment fields for the document
container file shall be Unicode strings encoded using the UTF-8
encoding scheme as specified in ISO/IEC 10646:2014, Annex D.

So a couple of cases:

1. A GEDCOM file contains a web-accessible file with an https url with a non-ascii (i.e., percent-encoded) character in it. Since it's web-accessible, the encoding is up to the hosting system.
2. A GEDZIP file contains a media file with a non-ascii name that is referenced from the gedcom.ged file. Since this is a local reference within a file format we define, I'd recommend we add a note to the Section 2.12 about case 2 saying GEDZIP _must_ use UTF-8 in file names.

dthaler:
Fixes #631


specification/gedcom-4-gedzip.md
```diff
@@ -14,7 +14,9 @@ Each GEDZIP file contains the following entries:
 - An entry for each *local file* `g7:type-FilePath` payload in `gedcom.ged`, with the same zip *file name* as the payload.
     If there is a local file named `gedcom.ged`, it must be renamed to a new unused filename with the same extension prior to constructing the GEDZIP.
 
-All file names inside a GEDZIP are case-sensitive.
+All file names inside a GEDZIP are case-sensitive UTF-8. Whereas a `g7:type-FilePath` [payload](#file-path)
+must percent-escape characters in the file name that cannot appear literally in a URI reference, zip *file names*
```
dthaler, line 18:
```suggestion
must percent-escape characters in the file name that cannot appear literally in a URI reference or valid URL string, zip *file names*
```

dthaler:

PR #642, opened 2025-05-06
Add example of PHRASE with non-OTHER enum

tychonievich:
As noted in #638, we have multiple PHRASE examples but none paired the PHRASE with a non-OTHER enumeration; this PR adds such an example.

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -1137,7 +1137,7 @@ A name given to a foundling orphan might be
 :::
 
 :::example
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -1136,6 +1136,16 @@ A name given to a foundling orphan might be
 ````
 :::
 
+:::example
+A record specificying a performer's "stage name" (a type of professional name) might become
```
tychonievich, line 1140:
```suggestion
A record specifying a performer's "stage name" (a type of professional name) might become
```

PR #651, opened 2025-05-23
Clarify FilePath syntax regarding non-ASCII characters

### LINKED ISSUE #648: Does RFC 3986 reference provide any real value?
https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#file-path says: "the payload is a URI reference as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), or a valid URL string as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/)"

RFC 3986 requires pct-encoding non-ascii characters in URI references (see the pchar production).
The WHATWG URL specification requires valid URL strings to have URL code points (see https://url.spec.whatwg.org/#url-code-points):

The URL code points are [ASCII alphanumeric](https://infra.spec.whatwg.org/#ascii-alphanumeric), U+0021 (!), U+0024 ($), U+0026 (&), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+003F (?), U+0040 (@), U+005F (_), U+007E (~), and [code points](https://infra.spec.whatwg.org/#code-point) in the range U+00A0 to U+10FFFD, inclusive, excluding [surrogates](https://infra.spec.whatwg.org/#surrogate) and [noncharacters](https://infra.spec.whatwg.org/#noncharacter).

So "naïve" uses a non-ascii character (U+00EF) which is in the range U+00A0 to U+10FFFD, inclusive.
This, it is not allowed in a URI reference but is allowed in a "valid URL string".

Because the FamilySearch GEDCOM 7 specification says "or", I think it means it is valid. This means that any application that only allows URI references could fail with strings valid under the "valid URL string" definition. If all URI references are valid URL strings, then does the RFC 3986 reference provide any real value?

dthaler:
Fixes #648

Defines the payload as just being a "valid URL string" (not "or URI reference") since the former includes the latter.

Explicitly stated that URLs can contain non-ascii characters as permitted in a "valid URL string".

Moved RFC 3986 link down to the bullet that already constrained local (but not remote) file paths to be URI references.

Explicitly stated that a local path requires percent-encoding non-ASCII characters, since the spec already required that to be the case.

### DIFF
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -339,8 +339,8 @@ Special = Text
 ## File Path
 
 The file path data type describes where an digital file is located in a machine-readable way.
-Syntactically, the payload is a URI reference as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), or a valid URL string as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).
-That is, it can be an absolute or relative URL, optionally with a fragment string.
+Syntactically, the payload is a "valid URL string" as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).
+That is, it can be an absolute or relative URL, optionally with a fragment string, and can contain non-ASCII characters that are permitted in a valid URL string.
 
 Version 7.0 only supports the following URLs:
 
@@ -350,14 +350,14 @@ Version 7.0 only supports the following URLs:
     and should be avoided in datasets that are expected to be shared on the web or with unknown parties,
     but may be appropriate for close collaboration between parties with known similar file structures.
 
-- A URI reference with all of the following:
+- A URI reference, as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), with all of the following:
     - no scheme
     - not beginning with `/` (U+002F)
     - not containing any path segments equal to `..` (U+002E U+002E)
     - not containing a reverse solidus character (U+005C `\`) or `banned` character, either directly or in escaped form
     - no query or fragment
     
-    refers to a **local file**. If the dataset is part of a [GEDZIP file](#gedzip), the URL of the local file is a zip archive filename; otherwise, the URL of a local file is resolved with *base* equal to the directory containing the dataset.
+    refers to a **local file**. If the dataset is part of a [GEDZIP file](#gedzip), the URL of the local file is a zip archive filename; otherwise, the URL of a local file is resolved with *base* equal to the directory containing the dataset.  Note that a URI reference cannot contain non-ASCII characters directly but instead percent-encodes them.
     
     It is recommended that local files use the directory prefix `media/`, but doing so is not required.
 


specification/gedcom-2-data-types.md
```diff
@@ -339,8 +339,8 @@ Special = Text
 ## File Path
 
 The file path data type describes where an digital file is located in a machine-readable way.
-Syntactically, the payload is a URI reference as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), or a valid URL string as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).
-That is, it can be an absolute or relative URL, optionally with a fragment string.
+Syntactically, the payload is a "valid URL string" as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/).
+That is, it can be an absolute or relative URL, optionally with a fragment string, and can contain non-ASCII characters that are permitted in a valid URL string.
```
tychonievich, line 343:
```suggestion
That is, it can be an absolute or relative URL, optionally with a fragment string, and can contain non-ASCII characters that are permitted in a valid URL string; it can also contain characters represented as percent-encoded UTF-8 bytes.
```

If we mention percent encoding elsewhere, I think we should here too.

specification/gedcom-2-data-types.md
```diff
@@ -350,14 +350,14 @@ Version 7.0 only supports the following URLs:
...
-- A URI reference with all of the following:
+- A URI reference, as defined by [RFC 3986](https://www.rfc-editor.org/info/rfc3986), with all of the following:
     - no scheme
     - not beginning with `/` (U+002F)
     - not containing any path segments equal to `..` (U+002E U+002E)
     - not containing a reverse solidus character (U+005C `\`) or `banned` character, either directly or in escaped form
     - no query or fragment
     
-    refers to a **local file**. If the dataset is part of a [GEDZIP file](#gedzip), the URL of the local file is a zip archive filename; otherwise, the URL of a local file is resolved with *base* equal to the directory containing the dataset.
+    refers to a **local file**. If the dataset is part of a [GEDZIP file](#gedzip), the URL of the local file is a zip archive filename; otherwise, the URL of a local file is resolved with *base* equal to the directory containing the dataset.  Note that a URI reference cannot contain non-ASCII characters directly but instead percent-encodes them.
```
tychonievich, line 360: I disagree with adding the percent-encoded requirement here. While I can see why it could be argued to be technically required by the previous text, I do no think it was an obvious reading of text that explicitly allowed WHATWG URLs too. Could we add something like "or a URL Path as defined by the [WHATWG URL specification](https://url.spec.whatwg.org/)" to not add this constraint on paths? (is that the correct production? I don't know WHATWG as well as I do RFC 3986)

tychonievich (changes requested):

specification/gedcom-2-data-types.md
dthaler: A "valid URL string" can also contain percent-encoded bytes that are not percent-encoded UTF-8 bytes, so I wonder if we should use different wording?  https://url.spec.whatwg.org/#url-units says "The URL units are [URL code points](https://url.spec.whatwg.org/#url-code-points) and [percent-encoded bytes](https://url.spec.whatwg.org/#percent-encoded-byte)."

(i.e.: URL code points are Unicode characters, but percent-encoded bytes need not be)

dthaler: Updated text.

PR #673, opened 2025-08-06
Fix paths and fragment URIs on gedcom.io

### LINKED ISSUE #665: Wrong extracted YAML files for type-DATE and type-LIST
The file `extracted-files/tags/type-Date` and <https://gedcom.io/terms/v7/type-Date> defines <https://gedcom.io/terms/v7/type-Date#exact> and <https://gedcom.io/terms/v7/type-Date#period> as well, but only has `uri: https://gedcom.io/terms/v7/type-Date` not the other two.

The file `extracted-files/tags/type-List` and <https://gedcom.io/terms/v7/type-List> defines <https://gedcom.io/terms/v7/type-List#Text> and <https://gedcom.io/terms/v7/type-List#Enum>, but has `uri: https://gedcom.io/terms/v7/type-List` (which is not the URI of a valid datatype).

This is an error in the current implementation of `tools/uri-def.py`. I know how to fix that, so I'm assigning this issue to myself.

In addition, the ideal fix would also requires change how Jekyl hosts these pages on <https://gedcom.io/>. But the fragment is not provided to the server when the page is visited, so I can see two options there:

1. We add JavaScript to the page, causing it to render correctly in most browsers.
2. We use the fact that YAML can have multiple documents in one file to put all the options.

Option 2 would look something like the following, which will be read fine by YAML libraries but looks strange to humans.

```yaml
%YAML 1.2
---
lang: en-US

type: data type

uri: https://gedcom.io/terms/v7/type-List#Enum

specification:
  - ...

---

lang: en-US

type: data type

uri: https://gedcom.io/terms/v7/type-List#Text

specification:
  - ...
...
```

### LINKED ISSUE #669: Make `push_to_gedcomio.py` support multiple versions, content type
URIs of terms in the GEDCOM specification (such as <https://gedcom.io/terms/v7/ADDR>) can be visited in a browser to find information about each term. This is implemented by `build/push_to_gedcomio.py`, which has several weaknesses.

- [ ] Should pick the URL (in the `permalink` and `redirect-from` metadata fields) from the URI field of the YAML file, not hard-code `/terms/v7/` for all files. This change will enable boy 5.5.1 and 7.1 files to also be hosted there.

- [ ] Should host valid pages, not the quasi-HTML it does now. Either it should provide a custom Content-Type HTTP header and stop wrapping the YAML in a div, pre, and code element; or it should include the rest of a valid HTML page (head, title, and so on).

     I don't see a way to pick a Content-Type in [the github pages docs](https://docs.github.com/en/contributing/writing-for-github-docs/using-yaml-frontmatter) but [Jekyll might provide it](https://jekyllrb.com/docs/configuration/webrick/), and it might be possible with a [custom theme](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/adding-a-theme-to-your-github-pages-site-using-jekyll).

tychonievich:
Resolves #665
Resolves #669 (at least the part the steering committee identified as a priority)

### DIFF
--- a/build/push_to_gedcomio.py
+++ b/build/push_to_gedcomio.py
@@ -102,7 +102,7 @@ for (path, [ts,txt]) in content.items():
     
 
 
-# also update the specification and changelong
+# also update the specification and changelog
 
 pdf = join(doc,'gedcom.pdf'), join(dst,'../../specifications','FamilySearchGEDCOMv7.pdf')
 html = join(doc,'gedcom.html'), join(dst,'../../specifications','FamilySearchGEDCOMv7.html')
--- a/build/push_to_gedcomio.py
+++ b/build/push_to_gedcomio.py
@@ -3,6 +3,7 @@ from sys import argv, stderr
 from os.path import dirname, isdir, join, getmtime, basename, exists
 from os import makedirs, utime
 from shutil import copyfile
+import re
 
 def get_paths():
     """Parses command-line arguments, if present; else uses defaults"""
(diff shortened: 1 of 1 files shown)


dthaler:
Approved during meeting

PR #677, opened 2025-08-14
Explain varied use of ALIA

### LINKED ISSUE #675: Clarify how applications used ALIA today
From https://github.com/FamilySearch/GEDCOM/discussions/674#discussioncomment-14084555_ we have the following:

* `ALIA` currently has caveats about possibly changing, and also has here-identified challenges in that different users/applications interpret its meaning differently. That suggests that using `ALIA` to communicate between applications requires some knowledge of how the other application is using `ALIA`, which knowledge goes beyond the spec.
* The steering committee differs in if the 7.0 text for `ALIA` is incompatible with the 5.5.1 text.
    * Proposed solution to clarity: add the text "Some users add ALIA instead of merging records to represent possible or uncertain identities", possibly adding in other interpretations.
    * But this also leaves `ALIA` under-defined and hard to use between applications.
* This suggests action is needed:
    * in 7.0 we should acknowledge that `ALIA` does not translate well between applications
    * in 7.1 we should add new, clearer tools and deprecate `ALIA`

tychonievich:
Resolves #675

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -244,13 +244,14 @@ An `ALIA` pointer should not point to the superstructure of the `ALIA`.
 This specification does not define how to connect `INDI` records with `ALIA`.
 Some systems organize `ALIA` pointers to create a tree structure, with the root `INDI` record containing the composite view of all facts in the leaf `INDI` records.
 Others distribute events and attributes between `INDI` records mutually linked by symmetric pairs of `ALIA` pointers.
-A future version of this specification may adjust the definition of `ALIA`.
 
 `ALIA` is known to be used for different purposes by different users.
 Some users use `ALIA` for uncertain connections, tentatively linking records prior to confirming identities and merging them into a single record;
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -240,6 +240,12 @@ A single individual may have facts distributed across multiple individual record
 
 An `ALIA` pointer should not point to the superstructure of the `ALIA`.
 
+`ALIA` is known to be used for different purposes by different users.
```
dthaler, line 243: Text is good.  Should this paragraph be in `:::note` like 249-254 are, or not?  It's not obvious to me what the criteria is for choosing between note and non-note.

albertemmerich: I think: yes, we should put it the the note.

tychonievich: Text moved

dthaler (approved):

dthaler (approved):
Approved during GEDCOM Steering Committee meeting.

PR #679, opened 2025-08-15
Draft v8.0 TEMPLATE, STICKY, FLEX, ASSET, GROUP, PROOF, alongside V7

mother10:
This PR models new structures for GEDCOM V8, to try and take that to a higher level. It is based on TEMPLATEs and STICKYs. They describe the evidence in documents, on gravestones, in a census, and any other evidence "entity" that can hold genealogical information. It has 3 huge enumtables so it becomes a bit "Data driven". It is structured alongside GEDCOM 7.x (for now) to help users and developers get used to this new system. But before implementation the group might decide to totally disconnect it from V7.x and start a new direction. In that case changes have to be applied to make this draft work.

While designing, time and effort have been spend to make it extendable, reliable, easily changeable, logical, solid and easy for users and developers.

It needs polishing, refinements, help from the group to find missing ends. It has a huge extra file with examples, they are not there for fun (although some have some fun inside) but they are there as a testbed. I heavily used that to make sure things would fit.

But I now want to show it all, I can hardly wait.😁


Norwegian-Sardines:
>The standardized ENUM tables ensure interoperability. User interface translations are a separate layer that every genealogy application already handles for current GEDCOM tags.

Where in the "Standardized Enum Table" is the definition for "Main", "Bride", "Biological Father/Mother", "Adoptive Father/Mother", "Foster Father/Mother", "Witness", "God Father/Mother", "Step Brother/Sister/Mother/Father", "Half Brother/Sister", etc?

I agree that GEDCOM is generally a "conclusion based" transmission and it would be nice that "Evidence without a conclusion" had a better role in the transmission of data.  This is a topic that goes beyond current use.  But this evidence layer should not replace the conclusions made by the transmitting party as you indicated in your comment:

> It is structured alongside GEDCOM 7.x (for now) to help users and developers get used to this new system. 

If I have enough evidence that pertaining to the Father/Mother of a child I would want to conclude the same!

Norwegian-Sardines:
> > ROLES` tables:  Where in the "Standardized Enum Table" is the definition for....................
> 
> https://github.com/FamilySearch/GEDCOM/blob/3900bf2c9e7ab15ae396365d2a4faaa167cf8e3d/specification/gedcom-3-structures-4-enumerations-roles.md

Yes I've seen these.

Issue #687, opened 2025-08-22
Proposal: Allow the header to have general Notes.


mother10:
Since two of the programs I currently use allow users to not only have `NOTE`s inside GEDCOM, at places the current specification allows them.  

But also because more users can have the need to somehow add Notes to a tree, that do not belong to a record. For instance general remarks about the tree, or the way they created their tree, or other things, this proposal adds the following to the `HEAD`:

Instead of:

```gedcom
  +1 <<NOTE_STRUCTURE>>                    {0:1}
  +1 <<CHANGE_DATE>>                       {1:1}
  +1 <<CREATION_DATE>>                     {1:1}
```
at the end of the `HEAD` it proposes:

```gedcom
  +1 <<NOTE_STRUCTURE>>                    {1:M}
  +1 <<CHANGE_DATE>>                       {1:1}
  +1 <<CREATION_DATE>>                     {1:1}
```

If this should clutter the head too much other solutions are welcome, but allow users somehow to add these kind of notes.

If another solution is chosen, I suggest to still have `NOTE` in the header as {1:1}, to allow software to describe for instance a conversion from 1 GEDCOM version to another.  
Many years have passed before the release of GEDCOM 7, and many files will now be converted since that release. 

But until now there is no way to somehow describe that conversion process. We have Notes and `CREA` and `CHAN` now to somehow track changes, but a conversion, which to me is an important process, cannot leave any trails it happened, when it happened, what file was involved and possibly more things, anywhere in GEDCOM.  

Users feel the need to be able to carefully describe their choice for one `INDI` above another, but the whole conversion process happens a bit "in the dark".

Norwegian-Sardines:
Your justification for requiring a `HEAD.NOTE` is based on conversion to GEDCOM v7.  Why would I be required to have a `HEAD.NOTE` if I'm creating the tree or GEDCOM for the first time?  Software is allowed to create this payload, forcing them to enter something that is unknown in a fresh start does not make sense to me!  

I agree that software should allow a user to add general `HEAD.NOTE` entries (many do not at this time) but to require them to have a one is not what I would want.

mother10:
> Why would I be required to have a HEAD.NOTE if I'm creating the tree or GEDCOM for the first time? 

Because it is easier to make it {1:1} then to make extra rules saying, the first creation of a file is an exception.  

It is **not the user** that enters the info, it is software, and its very easy to let that fill the note with something like "first creation with software xx on yy etc".   

[...]

Issue #219, opened 2022-10-10
Additional NOTE_STRUCTURE and SHARED_NOTE_RECORD properties


cdhorn:
I would like to propose some extensions to the NOTE/SNOTE constructs.

Notes may be used for many different purposes in programs, and so multiple notes may be associated with an object for specific purposes and there may be a need to differentiate between them. It would therefore be useful to support a TYPE field for notes to help identify the content of the note.

Notes may also be used to hold narrative text or stories, which one may want to display in a timeline or report in sequence with other events and data.  For this purpose it would be useful to have both a DATE as well as TITLE associated with the note.

ungeahnt:
>  It would therefore be useful to support a TYPE field for notes to help identify the content of the note.

NOTE (and also SNOTE) allows to add the HTML MIME type, so you can save additional formated information in HTML:

https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#NOTE_STRUCTURE
```
n NOTE <Text>                              {1:1}  g7:NOTE
  +1 MIME <MediaType>                      {0:1}  g7:MIME
```
[g7:MIME](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#MIME):

> 
> 
> MIME (Media type) g7:MIME
> 
> Indicates the media type of the payload of the superstructure, as defined by [BCP 13](https://www.rfc-editor.org/info/bcp13).
> 
> As of version 7.0, only 2 media types are supported by this structure:
> 
>     text/plain shall be presented to the user as-is, preserving all spacing, line breaks, and so forth.
> 
>     text/html uses HTML tags to provide presentation information. Applications should support at least the following:
>         p and br elements for paragraphing and line breaks.
>         b, i, u, and s elements for bold, italic, underlined, and strike-through text (or corresponding display in other locales; see [HTML §4.5](https://html.spec.whatwg.org/multipage/text-level-semantics.html) for more).
>         sup and sub elements for super- and sub-script.
>         The 3 XML entities that appear in text: &amp;, &lt; &gt;. Note that &quote; and &apos; are only needed in attributes. Other entities should be represented as their respective Unicode characters instead.
> 
>     Supporting more of HTML is encouraged. Unsupported elements should be ignored during display.
> 

Is this helpful?

ungeahnt:
Beside this, may it would be useful also allow additional mime-types for notes/snotes:

So you can use it for simple formated text (markdown) or add a list of URLs:

[...]

PR #723, opened 2025-11-11
Draft NOTE.TYPE and HEAD.NOTE cardinality

tychonievich:
Based on #219 and #687, using a single NOTE type for all contexts. The table of enumerated values is probably not the right one, but there were far more proposed in #219 than I have any confidence in describing.


tychonievich:

- Applies To might make sense for something like "Life Sketch", though past issues suggested we might want to make that an `INDIVIDUAL_ATTRIBUTE` instead. That said, "Life Sketch" seems likely to have ended up in `NOTE` in the past.
- `OTHER`.`PHRASE` lets applications use the `PHRASE` as headers or other navigation information to help users navigate the set of notes.
- Naming this `NOTE`.`TYPE` if it's an enum matches the rare case of `NAME`.`TYPE` rather than the common text-payload `TYPE` naming. It might be better to call this `NOTE`.`KIND` instead.
- One nice future solution (8.0) would be to have *all* cases use KIND and OTHER PHRASE instead of a mix of enum and free-text TYPE. But for 7.1 we can't do that everywhere (it's not backwards compatible), though we can use KIND to move in that direction.
- One criterion for what enums to include would be an application decision point that depends on the value. If we can find those, we could/should add them to the table.
    - An example might be `LIFE_SKETCH` which is often displayed in a different place/way than general notes.
    - Another example is import notes which are generated during GEDCOM import to provide information about the import process itself. Similar technical notes might be part of merging files. These notes are unlike human-generated notes and UIs might wish to separate the two kinds of notes (possibly even preventing users from overwriting/editing the technical log notes). We might name this `IMPORT_LOG` or `TECH_NOTES` or `CHANGE_LOG` or the like.
    - Another example is `TODO` which some applications have a way of displaying in a report of things that need doing.
    - Some users wish to have `AI_GENERATED` flagged as separate from human-generated. That might suggests a `KIND <List:Enum> {0:1}` or `KIND <List:Enum> {0:M}` instead of `KIND <Enum> {0:1}`.
        - Because we can envision substructures of `KIND` (such as human text clarifying what variety of `TODO` it is), `{0:M}` might be preferable.
    - There is work to do to decide which other proposed `KIND`s are programmatically important.

To do:

- [ ] Change from `TYPE` to `KIND`.
- [ ] Fix the `g7.1:` prefix for (transitive) superstructures.
- [ ] Think more carefully about the programmatic value of each value in the table, and explain that value in the table.

tychonievich:

- We know applications use: 
    - to do notes
    - data not present in other structures
    - life stories / memories
    - notes about (in)correctness of data
    - notes about ongoing research / future research directions
    - scopes for personas (what data goes in which INDI record)

    We may remove the others (ATTRIBUTION, SUMMARY, CONTEXT) from the PR because we are not aware of applications that use those.

- We know of applications that allow the user to chose which note types to export and/or display in a report. We discussed whether multiple OTHERs with different recommended visibility would help with that, but did not know what specific set of OTHER-like values to include and decided that asking the user what OTHER PHRASEs to show would make more sense.
   - Early programs such as PAF and AncestralQuest used a character such as a `!` or `~` to indicate that notes should be omitted from reports; that usage was never expressed in a specification, not uniformly supported by other application, and NOTE.TYPE (or KIND) is more appropriate for that situation.

specification/gedcom-3-structures-4-enumerations.md
```diff
@@ -67,6 +67,20 @@ except for tags used with different URIs for `FAM` vs `INDI`:
...
 
+
+### `g7.1:enumset-NOTE-KIND`
+
+| Value         | URI                  | Meaning                           |
+| :------------ | :------------------- | :-------------------------------- |
+| `DATA`        | `g7.1:enum-DATA`     | Historical data that is not present in other structures |
+| `RESEARCH`    | `g7.1:enum-RESEARCH` | Notes about research performed |
+| `REASONING`   | `g7.1:enum-REASONING`| Descriptions of why the data is (or is not) believed to be accurate |
+| `SCOPE`       | `g7.1:enum-SCOPE`    | A description of what data is or should be included in the given dataset (if `HEAD`.`NOTE`), persona (if `INDI`.`NOTE`) or the like |
```
tychonievich, line 78:
```suggestion
```

The members of the steering committee present in our meeting today could not remember where the `SCOPE` case was used in current applications with a meaning separate from `RESEARCH`.

dthaler:
Discussion in steering committee:
* We don't know of existing uses equivalent to SCOPE, so remove that one pending evidence of existing use, and then ok to merge.

PR #722, opened 2025-11-11
Add `BIRT`.`KIND BORN_DEAD` in v7.1

tychonievich:
This is part of the actions recommended in #106

PR #729, opened 2026-01-30
Redesign of repo tooling for mixed URI prefixes in 7.1

### LINKED ISSUE #727: URIs of Enumeration values in v7.1
There are a few places in the 7.0 specification and associated tooling that URIs are overly specific and don't adapt to 7.1:

- Specification:
    - [ ] Section 3.4's introductory paragraph says "each enumeration value defined in this section has a URI constructed by concatenating `g7:enum-` to the enumeration value". Currently there is no way to indicate when a specific enumeration value should use `g7.1:enum-` instead
    - [ ] Section 3.4 under `g7:enumset-EVEN` and `g7:enumset-EVENATTR` it says "these are given generic definitions with URIs constructed by concatenating `g7:enum-` to the enumeration value." It is not clear to me if these would need new URIs even if, for example, the CENS event were updated and given URI `g7.1:INDI-CENS` instead of `g7:INDI-CENS`.
    - [ ] Section 6.1 says "Each month defined in this section has a URI constructed by concatenating `g7:month-` to the standard tag"; that will stop being true if we add a new calendar in v7.1.

- Tooling
    - [ ] `uri-def.py` function `find_cat_tables` (around line 56) depends on the "by concatenating" language found in the specification, and may need updating if that wording changes.
    - [ ] `uri-def.py` lines 303 and 305 only look for `g7:enumset-`, not `g7.1:enumset-`.

tychonievich:
Closes #727
Unblocks #722 and #723

There's a lot going on here, and I spent several weeks trying to minimize this PR, split it into pieces, etc, but couldn't find a way that worked. So here's an outline of what's changed.

1. In the spec, many tables (notably enumsets) now have an explicit URI entry.

    I made the URIs their own column, not the two-line cells used for URIs of events and attributes earlier, mostly because I thought we have space. We can refactor that into using `<br>` if we prefer.

2. I re-wrote `build/uri-def.py` from the ground up as `build/extract-yaml-tsv.py`.

    `uri-def.py` was a 50-line script I wrote that I slowly changed into a 500-line monstristy that I find almost impossible to read or maintain. And it had many assumptions about the spec's format not only in its code (that's inevitable when parsing human text into machine-readable files) but also in its design. Trying to add the more versatile URIs was such a headache I decided to use it as an excuse for a long-overdue re-write.

    `extract-yaml-tsv.py` uses classes to have the code mirror the files it is designed to produce. It uses type annotations (which pass `mypy`) to avoid various typing bugs. It tries to do one thing in one place instead of doing all things possible everywhere. It's 100 lines longer than `uri-def.py` but faster to run and easier to maintain.

    I spent some effort to match idiosyncrasies of `uri-def.py`, like when to use `strings`, `'strings'`, or `"strings"` in the emitted YAML; but sometimes dong that was not possible without replicating `uri-def.py`'s worse parts, so there are some trivial changes (notably in the order of lists) in the files it creates.

3. I (accidentally) fixed various bugs of omission. `uri-def.py` was sloppy in how it handled the textual explanations in sections that had either gedstruct rules with alternatives or tables. Fixing that with a more principled approach means adding extra `specification:` entries for around half of all YAML files.

4. I tried but failed to match the sorting of the various .tsv files from `uri-def.py`, which was based on some kind of sorting of structures before they were used to generate multiple lines each. I was tempted to scrap that and simply sort the final lines, which is easy and well-defined, but this commit does not do that, instead doing a half-hearted effort at mimicking the old sorting.

I did all this on 7.1 because that was the original motivation, but if we like it I can port it back to the 7.0 branch (either by adding explicit URIs to the 7.0 spec or by adding special-case handling of tag-to-URI) with relative ease.


extracted-files/tags/REPO
dthaler: If the payload URI (line 23) is a v7.1 URI, shouldn't the structure URI also be v7.1?

build/extract-yaml-tsv.py
```diff
@@ -0,0 +1,599 @@
...
+      ans.setdefault(sect,[]).append(entry)
+  return ans
+
+
+def all_uri_section_text(txt:str, pfx:dict[str,str], data:dict[str,Concept]) -> tuple[dict[str,str], dict[str,list[str]]]:
+  """Finds all sections that are headed by a URI or URI-paired values.
+  Adds them to the provided dict
+  and returns 
+  - a dict from datatype names (like List:Enum) to their URIs
+  - a dict from table lables "event" and "attribute" to enumeration set URIs (like g7:enumset-EVENATTR) that include them
```
dthaler:
```suggestion
  - a dict from table labels "event" and "attribute" to enumeration set URIs (like g7:enumset-EVENATTR) that include them
```

build/extract-yaml-tsv.py
```diff
@@ -0,0 +1,599 @@
...
+        data[uri].subsumes.append(uri.replace('/v7.1/', '/v7/'))
+
+  # step 12: write all the various files
+
+  # 12.a: YAML files
+  for uri in data:
+    with open(pathof[uri], 'w') as dst:
+      print(data[uri], file=dst)
+
+  # 12.b: superstructure substructure cadinality
```
dthaler:
```suggestion
  # 12.b: superstructure substructure cardinality
```

tychonievich:
I think I resolved the previous issues: YAML is extracted by a separate file from TSV; TSV could be extracted from registries too; and URI version checks consider substructures, payloads, enumeration sets, and enumeration values.

The checks meant I found and fixed *many* transitive URI version changes:

- We'd changed record-SUBM to v7.1 earlier
- which now changes SUBM that points to it
- which changes SNOTE that contains SUBM
- which changes all event and attribute types which allow SNOTE
- which changes record-FAM which had those events
- which changes FAMS pointers
- ... and so on for a few related chains of URI updates

Since this already touches so many files, I also standardized to simple sorting for TSV files and substructure/superstructure lists to make the scripts simpler and more future-proof.

extracted-files/tags/NO-DATE
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/NO-DATE
+uri: https://gedcom.io/terms/v7.1/NO-DATE
```
dthaler: question: why did this one need to change (i.e., be bumped to v7.1 namespace)?
If the only change is to have a v7.1 superstructure instead of a v7 superstructure, I don't think that requires a namespace change for this, correct?

extracted-files/tags/ANCI
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/ANCI
+uri: https://gedcom.io/terms/v7.1/ANCI
```
dthaler, line 7: Question: to confirm, the payload is a pointer to an updated 7.1 SUBM record, so just want to confirm that the rule is the URI does need to change?

extracted-files/tags/CREA
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/CREA
+uri: https://gedcom.io/terms/v7.1/CREA
```
dthaler: I don't think this URI should be changed

extracted-files/tags/DATA-EVEN-DATE
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/DATA-EVEN-DATE
+uri: https://gedcom.io/terms/v7.1/DATA-EVEN-DATE
```
dthaler: I don't think this should be changed

extracted-files/tags/enum-EVEN
```diff
@@ -4,16 +4,21 @@ lang: en-US
 
 type: enumeration
 
-uri: https://gedcom.io/terms/v7/enum-EVEN
+uri: https://gedcom.io/terms/v7.1/enum-EVEN
```
dthaler, line 7: I don't think this URI should need to change, correct?  Same question for other `enum-*` files.

extracted-files/tags/FAMC-ADOP
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/FAMC-ADOP
+uri: https://gedcom.io/terms/v7.1/FAMC-ADOP
```
dthaler: I don't think this URI needs to change

extracted-files/tags/FAMC-STAT
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/FAMC-STAT
+uri: https://gedcom.io/terms/v7.1/FAMC-STAT
```
dthaler: I don't think this URI needs to change

extracted-files/tags/NOTE-TRAN
```diff
@@ -4,15 +4,15 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/NOTE-TRAN
+uri: https://gedcom.io/terms/v7.1/NOTE-TRAN
```
dthaler: I don't think this URI needs to change

extracted-files/tags/PLAC-FORM
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/PLAC-FORM
+uri: https://gedcom.io/terms/v7.1/PLAC-FORM
```
dthaler: I don't think this URI needs to change

extracted-files/tags/PLAC-TRAN
```diff
@@ -4,7 +4,7 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/PLAC-TRAN
+uri: https://gedcom.io/terms/v7.1/PLAC-TRAN
```
dthaler: I don't think this URI needs to change

extracted-files/tags/SOUR-DATA
```diff
@@ -4,13 +4,13 @@ lang: en-US
 
 type: structure
 
-uri: https://gedcom.io/terms/v7/SOUR-DATA
+uri: https://gedcom.io/terms/v7.1/SOUR-DATA
```
dthaler: I don't think this URI needs to change

extracted-files/tags/CREA
tychonievich: Agreed

extracted-files/tags/DATA-EVEN-DATE
tychonievich: Agreed

extracted-files/tags/enum-EVEN
tychonievich: The definition of `g7:enum-EVEN` was "either `g7:INDI-EVEN` or `g7:FAM-EVEN`." Since both of alternatives have new URIs, I think `g7.1:enum-EVEN` should too. Admittedly the updated URI is not in machine-readable part of the spec of these, but it is present in them.

extracted-files/tags/ANCI
tychonievich: That is how I've been applying the rule. Pointing to a new type means having a new URI.

extracted-files/tags/FAMC-STAT
tychonievich: Agreed

extracted-files/tags/NOTE-TRAN
tychonievich: Agreed

extracted-files/tags/PLAC-FORM
tychonievich: Agreed

extracted-files/tags/PLAC-TRAN
tychonievich: Agreed

extracted-files/tags/SOUR-DATA
tychonievich: I think I changed this because the definition mentions a new URI ("see `g7.1:DATA`"), but I agree that's not normative and can be rolled back to v7

extracted-files/tags/ANCI
tychonievich: Discussed in steering committee: we can see the reason in both ways, but decided that updating to 7.1 is appropriate

extracted-files/tags/enum-EVEN
tychonievich: Discussed in steering committee:

- The text might in enumset-ENUMATTR be better seen as non-normative because there are ways that a source could reference a CENS without being about either INDI-CENS or FAM-CENS, such as a source that lists when each census was performed by some nation or agency. Hence, maybe don't change these URIs because they are non-normative.
- That might be more complicated for enumset-EVEN, which is used in NO and is thus referencing a specific structure type.
- Having two different CENS URIs (v7 for ENUMATTR, v7.1 for ENUM) would be confusing. As the only cost to changing the URIs is extra files in the registry, we might update these URIs for consistency and simplicity.

tychonievich:

- Approved for be merged
- We should update the tooling in 7.0 to match this revised tooling, which will make diffing the two versions and making future updates much easier

PR #739, opened 2026-03-24
Editorial nits

dthaler:
I asked copilot to review some of the files looking for editorial issues. This PR fixes the ones found in the files touched.

For example, "Latter-Day Saints" should be "Latter-day Saints", not just in the name of the Church.

https://newsroom.churchofjesuschrist.org/article/latter-day-saints-101 is an example of a Church article that uses the term by itself in several places.


dthaler:
Approved during steering committee

PR #759, opened 2026-06-02
Add HEAD.TITL and HEAD.DESC

### LINKED ISSUE #750: HEAD has no specific title (TITL) or description (DESC)
Just to be clear on my meta-data issue: The HEAD has existing meta-data, but no specific title (TITL) or description (DESC). I currently use several different meta-data schemes in my HTML and SVG output, so I need a correspondence in GEDCOM. I am not suggesting that GEDCOM actually use DC, or schema.org, or open graphics, or RDF, or any other scheme -- simply that it have a correspondence in its own standard. -- Tony Proctor

_Originally posted by @ACProctor in https://github.com/FamilySearch/GEDCOM/discussions/747#discussioncomment-16570363_

tychonievich:
Add a title and description to the entire dataset.

Also adds a new translation structure `g7.1:TEXT-TRAN` for text-only payloads such as these.

Resolves #750

### DIFF
--- a/specification/gedcom-3-structures-1-organization.md
+++ b/specification/gedcom-3-structures-1-organization.md
@@ -155,6 +155,14 @@ n HEAD                                     {1:1}  g7.1:HEAD
   +1 LANG <Language>                       {0:1}  g7:HEAD-LANG
   +1 PLAC                                  {0:1}  g7:HEAD-PLAC
      +2 FORM <List:Text>                   {1:1}  g7:HEAD-PLAC-FORM
+  +1 TITL <Text>                           {0:1}  g7.1:HEAD-TITL
+     +2 LANG <Language>                    {0:1}  g7:LANG
+     +2 TRAN <Text>                        {0:1}  g7.1:TEXT-TRAN
+        +3 LANG <Language>                 {1:1}  g7:LANG
+  +1 DESC <Text>                           {0:1}  g7.1:HEAD-DESC
+     +2 LANG <Language>                    {0:1}  g7:LANG
+     +2 TRAN <Text>                        {0:1}  g7.1:TEXT-TRAN
+        +3 LANG <Language>                 {1:1}  g7:LANG
   +1 <<NOTE_STRUCTURE>>                    {0:1}
 ```
 
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -481,6 +481,10 @@ The `DatePeriod` covered by the entire source; the period during which this sour
 An [Individual Event](#individual-events).
 See also `INDIVIDUAL_EVENT_STRUCTURE`.
 
+#### `DESC` (Description) `g7.1:HEAD-DESC`
+
+A description of the entire dataset.
+
 #### `DESI` (Descendant Interest) `g7.1:DESI`
 
 Indicates an interest in research to identify additional descendants of this individual.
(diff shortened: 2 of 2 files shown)


specification/gedcom-3-structures-1-organization.md
```diff
@@ -155,6 +155,14 @@ n HEAD                                     {1:1}  g7.1:HEAD
   +1 LANG <Language>                       {0:1}  g7:HEAD-LANG
   +1 PLAC                                  {0:1}  g7:HEAD-PLAC
      +2 FORM <List:Text>                   {1:1}  g7:HEAD-PLAC-FORM
+  +1 TITL <Text>                           {0:1}  g7.1:HEAD-TITL
+     +2 LANG <Language>                    {0:1}  g7:LANG
```
dthaler, line 159: When would HEAD.TITL.LANG be different from HEAD.LANG?  Is that really a useful use case?
I can ask the same about DESC.LANG.

specification/gedcom-3-structures-1-organization.md
```diff
@@ -155,6 +155,14 @@ n HEAD                                     {1:1}  g7.1:HEAD
...
+     +2 TRAN <Text>                        {0:1}  g7.1:TEXT-TRAN
+        +3 LANG <Language>                 {1:1}  g7:LANG
+  +1 DESC <Text>                           {0:1}  g7.1:HEAD-DESC
```
dthaler, line 162: Should DESC have a media type?  E.g., is it legal to put HTML tags in the DESC text?  Or markdown?  What if you want to bold or italicize something in the description?

specification/gedcom-3-structures-1-organization.md
tychonievich: My goal was to match Dublin Core, which allows translations of the title. If we have TRAN we should also have LANG, right?

specification/gedcom-3-structures-1-organization.md
tychonievich: My goal was to match Dublin Core, where the description is a string. I also envisioned this being shown in a file preview box or the like where most systems I use don't allow markup.

specification/gedcom-3-structures-1-organization.md
dthaler: `TRAN` is fine.  I'm asking why we need TITL.LANG when we have HEAD.LANG.  I.e., why not:

```
+1 TITL <Text>   <-- in the language of HEAD.LANG
    +2 TRAN <Text>
        +3 LANG <Language>
````

We already say that when no LANG appears it's in HEAD.LANG.
So I'm asking what the use case is for TITL.LANG != HEAD.LANG, and I can't think of one.

specification/gedcom-3-structures-1-organization.md
dthaler: Ok

specification/gedcom-3-structures-1-organization.md
albertemmerich: My use case: I get a GEDCOM file with French content, HEAD.LANG declares French as main language in the file. To add it to my collection, I add a TITL and a DESC to the HEAD. As my French in not that perfect, I do this in German. If we do not add TITL.LANG, DESC.LANG to the spec I have a clear discrepancy in between the content of TITL and DESC and the LANG defined in HEAD for the file. So I have to break the standard, as I will not write in French. I think users do not care much about any LANG content in HEAD, however they should have the chance to do it correctly. I think the alternative to let HEAD.TITL empty and only fill HEAD.TITL.TRAN in German is no good solution either...

dthaler (approved):

tychonievich:

tychonievich:
It would be helpful to add this to an FAQ for v7.1 when that is released

PR #772, opened 2026-06-23
Remove NOTE-KIND SCOPE

### LINKED ISSUE #770: NOTE.KIND SCOPE
Discussion in steering committee:
* We don't know of existing uses equivalent to SCOPE, so remove that one pending evidence of existing use, and then ok to merge.

_Originally posted by @dthaler in https://github.com/FamilySearch/GEDCOM/issues/723#issuecomment-4721473904_

PR #723 was merged with `SCOPE` in it.  Should it be removed?

dthaler:
Fixes #770


albertemmerich:
I am not ready with the enum values for ```HEAD.NOTE.KIND```
Before we remove ```SCOPE``` from the list we should have some more discussion for the special situation of ```HEAD.NOTE```.
Example: I have running research (master version), versions preparing an online publication, versions with selected data only for exchange with a colleague, versions only giving a backup of the data for a certain time, and versions for test purposes. Instead of mixing up these classification with detailed information about the data contained in the file, I would propose a separate enum value of ```HEAD.NOTE.KIND``` - that could be ```SCOPE```

PR #774, opened 2026-06-23
Add tooling to support cross-major-version subsumes

### LINKED ISSUE #757: Tracking `subsumes` across major version changes
In reviewing how we mark when a v7.1 structure `subsumes` a v7.0 structure, we noted that some (but not all) v5.5.1 structures are subsumed by v7.0 structures. Adding that information would require two steps:

1. Identifying the subsumes relationships (the automated process used for minor versions will not work for major versions that might include breaking changes).
2. Adding something to the `build/extract-yaml.py` tool so that the corresponding `subsumes` entries are added to the v7.0 YAML files.

Doing this will also be something we need to consider when moving from v7.x to v8.0.

tychonievich:
The initial `specification/subsumes.tsv` created by this PR contains just a single entry, `ABBR`, which has the same defintion (modulo things like CONC and line value length limits) in both 5.5.1 an 7.0. It is present as a placeholder to demonstrate that this tooling works, but will cause a change to the ABBR 7.0 YAML file so the choice of that specific placeholder should be reviewed to ensure we agree that it is truly a subsumption.

Resolve #757

### DIFF
--- a/build/extract-yaml.py
+++ b/build/extract-yaml.py
@@ -563,13 +563,18 @@ if __name__ == '__main__':
         assert sub.split('/')[4] <= v, f"Enumeration value {sub} incompatible with enumset {uri}"
     
 
-  # step 12: add subsumes for any URI that also exists in earlier minor version
+  # step 12: add subsumes for any URI that also exists in earlier minor version or in subsumes.tsv
   from subprocess import run
   for uri in data:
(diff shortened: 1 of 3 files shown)


dthaler (approved):

tychonievich:

PR #775, opened 2026-06-23
Add note about varying use of BEF/AFT

### LINKED ISSUE #773: Meaning of AFT and BEF
This issue tracks any changes to the specification as opposed to the FAQ (which is covered by https://github.com/FamilySearch/GEDCOM.io/issues/397):

> Discussion in steering committee meeting:
> * We could note that users have differing interpretations of AFT and BEF and applications should be prepared to handle such varying interpretations rather than assuming a particular interpretation.
> * The term "inclusive" in the spec now allows (on import) either interpretation to be valid without losing information.  Using "exclusive" would have made one of the interpretations invalid.
> * However, users who use the exclusive interpretation in entering data result in imprecise data in the gedcom file.
> * There may also be sanity-checking code to generate warnings that may incorrectly generate warnings.  E.g., are AFT 1945 and BEF 1945 in conflict (and to be warned against) or not.
> * Besides the FAQ, the 7.0 spec should be updated to not state it is inclusive, but to say there are varying interpretations. 

 _Originally posted by @dthaler in [#397](https://github.com/FamilySearch/GEDCOM.io/issues/397#issuecomment-4781916596)_

`BET`/`AND` were unambiguously inclusive in both 5.5.1 and 7.0 so is outside the scope of this issue.

tychonievich:
I'm not sure this is the right note, but hope it will be a useful starting point for conversation leading to the right note.

Closes #773

### DIFF
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -130,6 +130,21 @@ The grammar above allows for `date`s to be preceded by various words. The meanin
 |`CAL` *x* |*x* is calculated from other data.          |
 |`EST` *x* |Exact date unknown, but near *x*; and *x* is calculated from other data.|
 
+:::note
+`AFT` and `BEF` were introduced in 5.0 without definition,
+defined with the single words "after" and "before" in 5.3,
+and defined as "event happened before/after the given date" in 5.4 through 5.5.1.
+Those definitions suggest that `AFT 1850` would mean "1 JAN 1851 or later,"
+while under the 7.0 definition `AFT 1850` means "1 JAN 1850 or later."
+
+This change in definitions was made to align with common (but not universal) practice.
+Given a source dated 1850 that asserts something had happened in the past
+(which could mean earlier that same year or in an earlier year),
+some users encode that as `BEF 1850` and others as `BEF 1851`.
+Both user entries are consistent with the source under the 7.0 definition,
+while only the second is consistent with the source under the 5.x definition.
+:::
+
 Known calendars and tips for handling dual dating and extension calendars are given in [Appendix A: Calendars and Dates](#A-calendars).
 
 `DateValue` and `DatePeriod` payloads may also be the empty string if no suitable form is known but a substructure (such as a `PHRASE` or `TIME`) is desired.
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -137,7 +137,6 @@ and defined as "event happened before/after the given date" in 5.4 through 5.5.1
 Those definitions suggest that `AFT 1850` would mean "1 JAN 1851 or later,"
 while under the 7.0 definition `AFT 1850` means "1 JAN 1850 or later."
 
-This change in definitions was made to align with common (but not universal) practice.
 Given a source dated 1850 that asserts something had happened in the past
 (which could mean earlier that same year or in an earlier year),
 some users encode that as `BEF 1850` and others as `BEF 1851`.


specification/gedcom-2-data-types.md
```diff
@@ -130,6 +130,21 @@ The grammar above allows for `date`s to be preceded by various words. The meanin
...
+and defined as "event happened before/after the given date" in 5.4 through 5.5.1.
+Those definitions suggest that `AFT 1850` would mean "1 JAN 1851 or later,"
+while under the 7.0 definition `AFT 1850` means "1 JAN 1850 or later."
+
+This change in definitions was made to align with common (but not universal) practice.
+Given a source dated 1850 that asserts something had happened in the past
+(which could mean earlier that same year or in an earlier year),
+some users encode that as `BEF 1850` and others as `BEF 1851`.
+Both user entries are consistent with the source under the 7.0 definition,
+while only the second is consistent with the source under the 5.x definition.
```
tychonievich, line 144:
```suggestion
These definitions left ambiguous whether they meant the beginning or end of the date in question:
did `BEF 1850` mean "before the beginning of 1850: 31 DEC 1849 or before"
or did it mean "before the end of 1850: 31 DEC 1850 or before"?
Some users found the inclusive definition more natural,
and would encode a source dated 1850 asserting something had happened in the past
as `BEF 1850`;
others preferred the more natural linguistic form of the exclusive definition,
and entered the same source as `BEF 1851`
because they didn't know if it happened earlier in 1850 or an earlier year.
Choosing one over the other also seemed to vary by region,
with some dialects and translations interpreting "before" and "after" differently than others.

The updated definition text in 7.0 is intended to be unambiguously inclusive:
`BEF 1850` could mean sometime within 1850, or before that date began.
The inclusive definition was chosen because it does not contradict the exclusive definition:
interpreting exclusive intent as inclusive data loses precision,
while interpreting inclusive intend as exclusive asserts falsehoods.
```

This alternative wording points out that there was an actual ambiguity, not just a sloppiness of use, in the 5.x wording, but also doesn't suggest that disagreements still exist. I can't decide if I like this text better or not.

tychonievich: Discussed in steering committee

- The first version is shorter and probably better than the second version
- Copying exclusive entry to exclusive meaning loses precision, but does not assert a falsehood
- The ambiguity is less extreme with more precise dates (`AFT 1 JUN 1850` has only 24 hours of ambiguity, while `AFT 1850` has 365 days of ambiguity)
- There may be some language difference, where English and German (two languages spoken by steering committee members) having different likelihoods of accepting the inclusive definition.
- We could deprecate using `AFT`/`BEF` with imprecise dates -- this could help rectify some of the ambiguity. An automated tool could turn all `AFT 1850` to either `AFT 31 DEC 1850` or `AFT 1 JAN 1850` depending on the culture/interpretation selected by the user.

specification/gedcom-2-data-types.md
```diff
@@ -130,6 +130,21 @@ The grammar above allows for `date`s to be preceded by various words. The meanin
...
 |`EST` *x* |Exact date unknown, but near *x*; and *x* is calculated from other data.|
 
+:::note
+`AFT` and `BEF` were introduced in 5.0 without definition,
+defined with the single words "after" and "before" in 5.3,
+and defined as "event happened before/after the given date" in 5.4 through 5.5.1.
+Those definitions suggest that `AFT 1850` would mean "1 JAN 1851 or later,"
+while under the 7.0 definition `AFT 1850` means "1 JAN 1850 or later."
+
+This change in definitions was made to align with common (but not universal) practice.
```
dthaler:
```suggestion
```

dthaler:
Approved during gedcom steering committee.
https://github.com/FamilySearch/GEDCOM.io/issues/397 tracks addition to the migration FAQ.
Discussion on what to do to fix the issue going forward is tracked by #778

lynchrs3:
Approved

lynchrs3 (approved):

PR #780, opened 2026-07-14
Draft note about metadata in FILEs

### LINKED ISSUE #779: Add note about Metadata in Media Files
### Discussed in https://github.com/FamilySearch/GEDCOM/discussions/777

<div type='discussions-op-text'>

<sup>Originally posted by **bawiseman** July  8, 2026</sup>
Gedcom 7 zip files can include media files.  Those files may contain metadata -- like face tags, title, description, date, location.  FamilySearch has been promoting a set of metadata guidelines around handling embedded metadata in family history photos.  See  [Photo Metadata](https://www.familysearch.org/innovate/photo-metadata).  Sharing the story of the photo can be much simpler with metadata.  Does Gedcom 7 have any guidelines or recommendations on how to handle embedded family history metadata in media files so that you would not have to add the metadata to the gedcom file directly, but read it out of the media file in a standard way (such as using the guidelines FamilySearch and other industry friends have recommended).</div>

tychonievich:
Resolve #779

I'm not sure that this is the right text; edits welcome.

I chose to make most of this addition normative because I think we want to make recommendations, not just alert readers to a topic of possible interest.

### DIFF
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -663,6 +663,22 @@ Applications are encouraged to show the metadata in referenced external files to
 and provide ways of integrating that metadata into appropriate structures with user guidance.
 
 :::note
+Conversion between file metadata and GEDCOM structures generally requires user input, as the mapping is often content-dependent.
+For example, the if `image.jpg` has a `dc:title` field and is used in the following way
+
+```gedcom
+0 @I1@ INDI
+1 SOUR @S1@
+0 @S1@ SOUR
+1 OBJE @O1@
+0 @O1@ OBJE
+1 FILE image.jpg
+```
+
+that field might belong in a `NAME` of `@I1@`, or a `TITL` in `@S1@`, or a `NOTE` in `@O1@`, or possible other locations depending on what the field contains.
+:::
+
+:::note
 Information about embedded metadata of interest to family history can be found at <https://www.familysearch.org/innovate/photo-metadata>.
 :::
 
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -658,7 +658,7 @@ See the [File Path datatype](#file-path) for more details.
 
 External files may contain embedded metadata,
 which may be of value to family history work.
-Applications should not to remove or edit metadata except as requested by the user.
+Applications should not remove or edit metadata except as requested by the user.
 Applications are encouraged to show the metadata in referenced external files to the user
 and provide ways of integrating that metadata into appropriate structures with user guidance.
 
--- a/specification/gedcom-3-structures-3-meaning.md
+++ b/specification/gedcom-3-structures-3-meaning.md
@@ -668,7 +668,7 @@ and provide ways of integrating that metadata into appropriate structures with u
 
 :::note
 Conversion between file metadata and GEDCOM structures generally requires user input, as the mapping is often content-dependent.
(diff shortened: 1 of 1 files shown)


specification/gedcom-3-structures-3-meaning.md
```diff
@@ -656,6 +656,16 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
 A reference to an external file.
 See the [File Path datatype](#file-path) for more details.
 
+External files may contain embedded metadata,
+which may be of value to family history work.
+Applications should not to remove or edit metadata except as requested by the user.
```
dthaler:
```suggestion
Applications should not remove or edit metadata except as requested by the user.
```

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -656,6 +656,16 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
...
+Applications are encouraged to show the metadata in referenced external files to the user
+and provide ways of integrating that metadata into appropriate structures with user guidance.
```
dthaler, line 663: Can we say anything more about what is "appropriate"?  E.g., could we have a table of metadata item <-> appropriate GEDCOM structure?

tychonievich: I don't think we realistically can. There are hundreds of metadata tags defined in multiple standards and some evidence that common applications have their own *de facto* standard way of using tags that differs from those standards.

Even were that not the case, it's still not clear where data goes. Consider the `dc:title` metadata field of `some_image.jpg` used in the following way:

```gedcom
0 @I1@ INDI
1 SOUR @S1@
0 @S1@ SOUR
1 OBJE @O1@
0 @O1@ OBJE
1 FILE some_image.jpg
```

That `dc:title` might belong in any of the following places:

```gedcom
0 @I1@ INDI
1 SOUR @S1@
1 NAME maybe goes here (1)
1 NOTE maybe goes here (2)
0 @S1@ SOUR
1 OBJE @O1@
2 TITL maybe goes here (3)
1 TITL maybe goes here (4)
1 NOTE maybe goes here (5)
0 @O1@ OBJE
1 FILE some_image.jpg
2 TITL maybe goes here (6)
1 NOTE maybe goes here (7)
```

I don't think an application can pick a correct location without human help, hence the "with user guidance" part of the proposed text.

dthaler: Thanks.  I think explaining that in a note (maybe using a subset of the example that shows 2 or 3 places something could go) would help the reader understand what "with user guidance" means.

dthaler (approved):

specification/gedcom-3-structures-3-meaning.md
```diff
@@ -660,6 +660,32 @@ See also `INDIVIDUAL_EVENT_STRUCTURE`.
...
 
+External files may contain embedded metadata,
+which may be of value to family history work.
+Applications should not remove or edit metadata except as requested by the user.
+Applications are encouraged to show the metadata in referenced external files to the user
+and provide ways of integrating that metadata into appropriate structures with user guidance.
+
+:::note
+Conversion between file metadata and GEDCOM structures generally requires user input, as the mapping is often content-dependent.
+For example, the if `image.jpg` has a `dc:title` field and is used in the following way
```
tychonievich:
```suggestion
For example, if `image.jpg` has a `dc:title` field and is used in the following way
```

dthaler (approved):

Issue #35, opened 2021-08-12
Add support for Islamic calendars


elyoh:
This proposal formally defines an abstract Islamic calendar, which can be used to represent various in use forms of the Islamic calendar (as discussed in #1 ).

`ISLAMIC`

The Islamic calendar is a deterministic lunar Hijri calendar, based on a 30-year cycle.  There are two possible epochs for this calendar.  In the civil epoch, 1 Muḥarram 1 aligns with Julian day 16 July 622.  In the astronomical epoch, 1 Muḥarram 1 aligns with Julian day 15 July 622.

Months have fixed lengths of 29 or 30 days, except for the twelth month, Dhū al-Ḥijja, which has an extra day in a leap year.

Permitted months are:
Code | Name | Days
-- | -- | --
MUHAR | Muḥarram (المُحَرَّم) | 30
SAFAR | Ṣafar (صَفَر) | 29
RABIA | Rabī I (رَبيع   الاوَّل) | 30
RABIT | Rabī II (رَبيع الآخِر) | 29
JUMAA | Jumādā I (جُمادى الأولى) | 30
JUMAT | Jumādā II (جُمادى الآخِرة) | 29
RAJAB | Rajab (رَجَب) | 30
SHAAB | Sha‘bān (شَعبان) | 29
RAMAD | Ramaḍān (رَمَضان) | 30
SHAWW | Shawwāl (شَوّال) | 29
DHUAQ | Dhū al-Qa‘da (ذو القَعدة) | 30
DHUAH | Dhū al-Ḥijja (ذو الحِجّة) | 29 (or 30 in a leap year) 

In every 30-year cycle, 11 years are leap years.  The determination of leap years depends on the variant in use.

The epoch marker `BH` (before Hijra) is permitted in this calendar; year y BH indicates a year y years before year 1. Thus, there is no year 0; year 1 BH was followed by year 1.

**_Note on variants_**

The determination of leap years follows one of four rules.

Variant | Leap years with 355 days in each 30-year cycle
-- | --
I | 2, 5, 7, 10, 13, 15, 18, 21, 24, 26, 29
II | 2, 5, 7, 10, 13, 16, 18, 21, 24, 26, 29
III | 2, 5, 8, 10, 13, 16, 19, 21, 24, 27, 29
IV | 2, 5, 8, 11, 13, 16, 19, 21, 24, 26, 30

**_Note on epochs_**
There are two possible epoch dates for each variant.
[...]

PR #798, opened 2026-08-19
Allow reusing standard months in extension calendars

### LINKED ISSUE #797: Extension calendars should not require extension tags for months that already have standard tags defined
> Yes please. If it were implemented as an extension, then (if rules are strictly applied) it must use `extTag` for all month names, which would be a pity.
> 
> https://github.com/FamilySearch/GEDCOM/blob/4fd1d1af98d2ca4986577b5cd1e148973f37d99c/specification/gedcom-2-data-types.md?plain=1#L113 

 _Originally posted by @BertKoor in [#28](https://github.com/FamilySearch/GEDCOM/issues/28#issuecomment-5264223638)_

dthaler:
Fixes #797

### DIFF
--- a/specification/gedcom-2-data-types.md
+++ b/specification/gedcom-2-data-types.md
@@ -110,7 +110,9 @@ In addition to the constraints above:
     All known calendars restrict `day` to be between 1 and a month-specific maximum.
     The largest known maximum is 36, and most months in most calendars have a lower maximum.
 - No calendar names, months, or epochs match `dateRestrict`.
-- Extension calendars (those with `extTag` for their `calendar`) must use `extTag`, not `stdTag`, for months.
+- Extension calendars (those with `extTag` for their `calendar`) must use `stdTag` for months
+  only when those months are already defined in a standard calendar.  Extension calendars must
+  use `extTag` for any other months defined.
 
 It is recommended that calendars avoid using a single tag to refer to both a month and an epoch.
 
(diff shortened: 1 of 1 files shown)


specification/gedcom-2-data-types.md
```diff
@@ -110,7 +110,9 @@ In addition to the constraints above:
     All known calendars restrict `day` to be between 1 and a month-specific maximum.
     The largest known maximum is 36, and most months in most calendars have a lower maximum.
 - No calendar names, months, or epochs match `dateRestrict`.
-- Extension calendars (those with `extTag` for their `calendar`) must use `extTag`, not `stdTag`, for months.
+- Extension calendars (those with `extTag` for their `calendar`) must use `stdTag` for months
+  only when those months are already defined in a standard calendar.  Extension calendars must
+  use `extTag` for any other months defined.
```
tychonievich: I worry that the "must use `stdTag`" part will make each new version invalidate older extensions. Suppose an extension defines `_SAFAR` (as suggested in #35) and then v7.1 adds that as a standard month, but with a different set of variants of the Islamic calendar than the extension. The extension is thus still needed because it covers other calendars, and while I'd encourage it to update to the standard tag I'd not think that it was required to do so.

```suggestion
- Extension calendars (those with `extTag` for their `calendar`) should use `stdTag` for months
  which are already defined in a standard calendar.  Extension calendars must
  use `extTag` for any other months defined.
```

tychonievich: Or we could remove this bullet entirely. I think this is already better covered in Appendix A, which reads in part:

> All month tags must either be standard tags defined for the month name in some standard calendar or be extension tags.

> Months with extension tags are permitted in standard calendars only when they are documented extension tags with standard URIs defined by the calendar. This is intended for future compatibility, to accommodate cases where an extension calendar later becomes standardized without changing the URI, at which point the calendar name and month names with `_` are acceptable with the now-standard URIs.

dthaler:
```suggestion
```

tychonievich (approved):

dthaler (commented):

tychonievich (approved):

